PR #78, opened 2015-02-12
Allow moment

princemaple:
:koala: 

/cc @zendesk/quokka


sandlerr:
+1

danielbreves:
?

svizzari:
:+1:

princemaple:
Public: we now have access to moment js in Zendesk agent interface, so apps can use moment.js without including it as an extra file.

ocke:
:+1: Great news guys. We use it all the time and now we can remove it from our apps lib dir.

PR #79, opened 2015-02-12
Clean up gemspec

princemaple:
:koala:


princemaple:
Ready now
@zendesk/quokka

zendesk_apps_support.gemspec
```diff
@@ -17,10 +17,10 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'json-stream'
   s.add_runtime_dependency 'image_size'
   s.add_runtime_dependency 'erubis'
-  s.add_runtime_dependency 'jshintrb',    '0.2.4'
+  s.add_runtime_dependency 'jshintrb'
```
sandlerr: why? We don't want people using older versions of jshintrb, that would fail some of our official apps.

princemaple:
@sandlerr right

sandlerr:
+1

atroche:
:+1:

PR #82, opened 2015-03-30
Swap out MultiJson in favor of JSON

princemaple:
:koala:
@sandlerr 

/cc @zendesk/quokka

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,4 @@
 source 'https://rubygems.org'
 
 gemspec
+gem 'byebug'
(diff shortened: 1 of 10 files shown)


sandlerr:
:+1: so we still depend on this because of `jshintrb`

Gemfile
```diff
@@ -1,3 +1,4 @@
 source 'https://rubygems.org'
 
 gemspec
+gem 'byebug'
```
princemaple: lets leave it there

atroche:
:+1:

PR #85, opened 2015-05-28
[WIP] Precompile Handlebars templates

svizzari:
:koala: :no_entry_sign: :warning: 

POC/WIP to precompile Handlebars templates on the server rather than in the client.

/cc @danielbreves 

### Tasks
- [ ] Investigate rendering out precompiled templates to a file, caching that file as an app asset
- [ ] Investigate correct function of custom Handlebars helpers
- [ ] Update tests

### References
- Jira link: 

### Risks
- [HIGH] Templates not rendering as expected

### DIFF
--- a/lib/zendesk_apps_support/assets/templates.js.erb
+++ b/lib/zendesk_apps_support/assets/templates.js.erb
@@ -1,0 +1,8 @@
+<% unless templates.nil? %>
+{
+  <% templates.each do |name, fn| %>
+  <%= name %>: <%= fn %>,
+  <% end %>
+  eom: undefined
+}
+<% end %>
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,6 +3,7 @@ PATH
   specs:
     zendesk_apps_support (1.18.0)
       erubis
+      execjs
       i18n
       image_size
       jshintrb (~> 0.2.4)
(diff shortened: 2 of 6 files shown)


lib/zendesk_apps_support/assets/templates.js.erb
```diff
@@ -0,0 +1,8 @@
+<% unless templates.nil? %>
+{
+  <% templates.each do |name, fn| %>
+  <%= name %>: <%= fn %>,
```
ocke: I had to do the following to make it work:
That also what the normal compiler outputs btw (it wraps it into that Handlebars.template function).
```
-  <%= name %>: <%= fn %>,
+  <%= name %>: Handlebars.template(<%= fn %>),
```

svizzari: Hey @ocke, yep

I only began testing when I first put it up, realised there was a call missing, then moved onto something else :smile_cat: 

Thanks for the comment, much appreciated

ocke: Well now you know I tried your code ;-). I actually programmed this thing myself, not realizing you did it already.

I used npm and `.bin/handlebars #{file}` though. But I like your approach more.

svizzari: You probably saw, in the PR description, I considered that, as an enhancement, it might be better to precompile the templates out to a file to be cached in S3, rather than precompiling every time.

PR #90, opened 2015-08-26
added version and app name for zas!

ocke:
Added the same three properties to app.js to mimic zam correctly.

Installation name is the same as the app name. In ZAM these are different.

https://github.com/zendesk/zendesk_app_market/pull/1378

/cc @zendesk/quokka


danielbreves:
:+1:

PR #97, opened 2015-12-09
Olaf/es6

ocke:
:koala:

to_es6 function for zas.

I used therubyracer because node was incredible slow!
- a request for plain js: 10-15ms
- a request with using (babel) es6 and node: 700 ms! (man it was so slow, really annoying)
- a request using babel and therubyracer: 20-25ms. (that's more like it!)

I needed to change app.js to be a module, as babel only support modules, which means the top line becomes: `module.exports = {` instead of `function() { return {`.

anyway have a play!

/cc @zendesk/quokka
### todo:
- DONE: backwards compatibility for app.js
### References
- Jira link: 
### Risks
- None


zendesk_apps_support.gemspec
```diff
@@ -17,6 +17,8 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'image_size'
   s.add_runtime_dependency 'erubis'
   s.add_runtime_dependency 'jshintrb', '~> 0.3.0'
+  s.add_runtime_dependency 'babel-transpiler'
+  s.add_runtime_dependency 'therubyracer'
```
sandlerr: doesn't babel pull this in automatically?

sandlerr:
this looks very slick. If we put it behind an undocumented manifest option I'd be pretty happy to ship it soon

adammw:
@ocke is there any way to make the rubyracer gem optional? some people may not want to deal with the hassle that is compiling and obtaining rubyracer for something they may not want to use.

svizzari:
Why is the method called `to_es6`?  It's transpiling from es2015 -> es5, right?

lib/zendesk_apps_support/assets/src.js.erb
```diff
@@ -1,16 +1,16 @@
...
+        <%= content %>
+      },
+    <% end %>
+    'app.js': function(exports, require, module) {
+      <%= source %>
+    }
+  };
 
-  var source = <%= source %>;
+  var source = require('app');
```
adammw: what if there's a module already called 'app' under the lib directory?

adammw:
I was going to suggest something like 'transpile' because it depends really on the options you give babel what it'll do

lib/zendesk_apps_support/assets/src.js.erb
svizzari: Generate a unique-ish key?

adammw: `_root`? it's all a bit hacky.

svizzari:
@ocke This PR would need to switch out using JSHint for ESLint or similar, if we were going to allow app developers to upload ES6 code, right?  Sure, JSHint has a harmony mode, but ESLint might be more appropriate?

zendesk_apps_support.gemspec
ocke: nope.

sandlerr: it depends on [execjs](https://github.com/babel/ruby-babel-transpiler/blob/master/babel-transpiler.gemspec#L22) if you want to not force people to install therubyracer then just go with that, and install therubyracer in our private consumer of this gem.

PR #100, opened 2015-12-23
Double app.js fix

ocke:
:koala:

app.js was being included in the modules because the tests appended app.js to package.@lib_files which was terrible, so I fixed it.

/cc @zendesk/quokka

### References
 - Jira link: 

### Risks
 - None

### DIFF
--- a/lib/zendesk_apps_support/package.rb
+++ b/lib/zendesk_apps_support/package.rb
@@ -77,11 +77,22 @@ module ZendeskAppsSupport
     end
 
     def files
-      non_tmp_files
+      files = []
+      Dir[root.join('**/**')].each do |f|
+        next unless File.file?(f)
+        relative_file_name = f.sub(/#{root}\/?/, '')
+        next if relative_file_name =~ /^tmp\//
+        files << AppFile.new(self, relative_file_name)
+      end
+      files
+    end
+
+    def js_files
+      files.select { |f| f =~ /\.js$/ }
     end
 
     def lib_files
-      @lib_files ||= non_tmp_files.select { |f| f =~ /^lib\/.*\.js$/ }
+      @lib_files ||= files.select { |f| f =~ /^lib\/.*\.js$/ }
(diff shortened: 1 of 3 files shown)


lib/zendesk_apps_support/package.rb
```diff
@@ -77,11 +77,22 @@ def locales
     end
 
     def files
-      non_tmp_files
+      files = []
+      Dir[root.join('**/**')].each do |f|
+        next unless File.file?(f)
+        relative_file_name = f.sub(/#{root}\/?/, '')
```
sandlerr: this probably doesn't need to be a regex does it?

lib/zendesk_apps_support/package.rb
```diff
@@ -77,11 +77,22 @@ def locales
...
+      files
+    end
+
+    def js_files
+      files.select { |f| f =~ /\.js$/ }
     end
 
     def lib_files
-      @lib_files ||= non_tmp_files.select { |f| f =~ /^lib\/.*\.js$/ }
+      @lib_files ||= files.select { |f| f =~ /^lib\/.*\.js$/ }
```
sandlerr: let's use `js_files` and only test if it starts with `lib`?

sandlerr:
:+1: on green

lib/zendesk_apps_support/package.rb
```diff
@@ -77,11 +77,22 @@ def locales
...
+      files
+    end
+
+    def js_files
+      @js_files ||= files.select { |f| f =~ /\.js$/ }
     end
 
     def lib_files
-      @lib_files ||= non_tmp_files.select { |f| f =~ /^lib\/.*\.js$/ }
+      @lib_files ||= js_files.select { |f| f =~ /^lib\/.*\.js$/ }
```
sandlerr: how that you're using `js_files` the regex can just be `/^lib\//` can't it?

PR #103, opened 2016-01-07
[AF-224] Add Location

danielbreves:
:koala:

to replace our Location model in ZAM

/cc @zendesk/quokka

### Risks
 - Apps with valid locations may fail validation
 - Apps with invalid locations may validate successfully

### DIFF
--- a/lib/zendesk_apps_support/location.rb
+++ b/lib/zendesk_apps_support/location.rb
@@ -21,7 +21,7 @@ module ZendeskAppsSupport
         LOCATIONS_AVAILABLE.keys
       end
 
-      def for(host)
+      def names_for(host: host)
         LOCATIONS_AVAILABLE[host.to_s].keys
       end
     end
--- a/lib/zendesk_apps_support/location.rb
+++ b/lib/zendesk_apps_support/location.rb
@@ -22,7 +22,7 @@ module ZendeskAppsSupport
       end
 
       def for(host)
-        LOCATIONS_AVAILABLE[host].keys
+        LOCATIONS_AVAILABLE[host.to_s].keys
       end
     end
   end
--- a/lib/zendesk_apps_support/location.rb
+++ b/lib/zendesk_apps_support/location.rb
@@ -1,0 +1,29 @@
+module ZendeskAppsSupport
+  module Location
+    # the numbers below match the enum values on the database, do not change them!
+    LOCATIONS_AVAILABLE = {
+      "zendesk" => {
+        "top_bar" => 1,
+        "nav_bar" => 2,
+        "ticket_sidebar" => 3,
+        "new_ticket_sidebar" => 4,
+        "user_sidebar" => 5,
+        "organization_sidebar" => 6,
+        "background" => 7
+      },
+      "zopim" => {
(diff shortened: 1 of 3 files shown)


lib/zendesk_apps_support/location.rb
```diff
@@ -0,0 +1,53 @@
...
+      }
+    }.freeze
+
+    class << self
+      def to_enum
+        @enum ||= LOCATIONS_AVAILABLE.values.reduce(&:merge)
+      end
+
+      def ids_for(host: :zendesk, locations: [])
+        host_app = LOCATIONS_AVAILABLE[host]
```
adammw: maybe host_locations or host_location_map?  (I keep thinking this is an app)

lib/zendesk_apps_support/location.rb
```diff
@@ -0,0 +1,29 @@
...
+        "chat_sidebar" => 8
+      }
+    }.freeze
+
+    class << self
+      def hosts
+        LOCATIONS_AVAILABLE.keys
+      end
+
+      def for(host)
```
sandlerr: should we handle the case where someone passes a symbol instead of a string?

sandlerr:
:+1:

sandlerr:
@zendesk/vegemite :+1:?

ocke:
:+1:

lib/zendesk_apps_support/location.rb
```diff
@@ -0,0 +1,29 @@
...
+        LOCATIONS_AVAILABLE[host.to_s].keys
```
adammw: This will return the location_ids right not "Location" objects? Perhaps `location_ids_for_host` is more appropriate?

lib/zendesk_apps_support/location.rb
```diff
@@ -0,0 +1,29 @@
+module ZendeskAppsSupport
+  module Location
+    # the numbers below match the enum values on the database, do not change them!
```
adammw: uh, is this something we want to publically have in our codebase? (i do realise the irony of asking this question in a public repo)

lib/zendesk_apps_support/location.rb
danielbreves: ZAS has no knowledge of the `Location` object. How about `ids_for`?

lib/zendesk_apps_support/location.rb
danielbreves: I wondered about that to, couldn't think of any problems with leaving it. We definitely need a warning of some sort.

lib/zendesk_apps_support/location.rb
adammw: Well, apart form that this class is it self defining one :) `ids_for` is okay, but maybe make host a named parameter so it's at least somewhat clear what arguments it takes? (or do we not have ruby 2.0 as a minimum requirement?)

adammw: sorry, i misread, this method returns the location names not ids right?

danielbreves:
@adammw :haircut:

adammw:
:+1:

PR #94, opened 2015-11-16
Standardize strings

danielbreves:
:koala:

Seems like we have been editing en.yml directly.

/cc @zendesk/quokka @ydrain @pdeuter 

### Risks
 - None

PR #107, opened 2016-01-23
HD-612 Strings: reverse sweep

pdeuter:
@craig-day @coreysf @mariasica @ydrain @svizzari 

### Risks
low.  English edits and title improvements.

### DIFF
--- a/config/locales/translations/zendesk_apps_support.yml
+++ b/config/locales/translations/zendesk_apps_support.yml
@@ -27,7 +27,7 @@ parts:
       key: "txt.apps.admin.error.app_build.no_app_js_required"
       title: "App builder job: ban app.js while requirements only"
       value: "Do not set your app to requirements only if you need app.js"
-      obsolete: 2016-02-28
+      obsolete: "2016-02-28"
   - translation:
       key: "txt.apps.admin.error.app_build.manifest_not_json"
       title: "App builder job: manifest is invalid JSON error"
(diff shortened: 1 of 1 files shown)


mariasica:
:+1:  on green

pdeuter:
@svizzari I am going to merge despite red as I don't think my changes are causing the failure.

svizzari:
:+1:

adammw:
The [one error on Travis](https://travis-ci.org/zendesk/zendesk_apps_support/jobs/104219377#L137) mentions something about `bundle exec rake i18n:standardize`, I'm pretty sure the failures *are* related to this PR as after the reverse sweep the translations aren't in sync with config/locales/en.yml (which is what I believe running that task does). 

That being said, I'm a little worried that the test is so flaky that it only failed once out of 4 ruby versions, plus didn't failed at all on the travis push hook.

svizzari:
It seems like the `bundle exec rake i18n:standardize` task should be run as a step in the reverse sweep PR, otherwise it isn't being run at all, right?  Looks like in the past we've manually run the task in another PR, after each reverse sweep, i.e. https://github.com/zendesk/zendesk_apps_support/pull/94.

It seems like we could almost run the standardize task every time we do a spec run since the file task will no-op if the zendesk i18n file isn't newer.  @adammw, thoughts?

svizzari:
Disregard above comment, both files need to be checked in. Task should be run with reverse sweep before merge, if translation values have changed.

craig-day:
As an aside, I don't think adding a one-off process for Paul during the reverse sweep is okay. Every week, he has to reverse sweep into every repo that has strings so adding a random step to 1of 50+ repositories is inconvenient to say the least. I'm just trying to make everyone aware of the process and maybe @pdeuter has a different opinion, but I just thought I would point it out.

sandlerr:
Is this not done by a script?

svizzari:
@craig-day Yep, the process needs to be as frictionless as possible, agreed.  Lets work out how we can make the process around this repo more robust without adding to @pdeuter workload.  My last comment was more to the effect that the *right time* to run the Rake task is with the reverse sweep PR.  @pdeuter, are you manually creating the reverse sweep PRs?

PR #108, opened 2016-02-01
AF-238 orderable attribute

sandlerr:
:v:

/cc @zendesk/vegemite

### Description

Add an orderable attribute to the locations. IMO this is a great candidate for replacement with a [Struct](http://ruby-doc.org/core-2.3.0/Struct.html) but they are a bit less flexible and people are less familiar with them.

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-238

### Risks
* malformed installed.js (apps don't load)
* breaks reordering

### DIFF
--- a/lib/zendesk_apps_support/location.rb
+++ b/lib/zendesk_apps_support/location.rb
@@ -3,16 +3,16 @@ module ZendeskAppsSupport
     # the numbers below match the enum values on the database, do not change them!
     LOCATIONS_AVAILABLE = {
       'zendesk' => {
-        'top_bar' => 1,
-        'nav_bar' => 2,
-        'ticket_sidebar' => 3,
-        'new_ticket_sidebar' => 4,
-        'user_sidebar' => 5,
-        'organization_sidebar' => 6,
-        'background' => 7
+        'top_bar' => { 'id' => 1, 'orderable' => true },
+        'nav_bar' => { 'id' => 2, 'orderable' => true },
+        'ticket_sidebar' => { 'id' => 3, 'orderable' => true },
+        'new_ticket_sidebar' => { 'id' => 4, 'orderable' => true },
+        'user_sidebar' => { 'id' => 5, 'orderable' => true },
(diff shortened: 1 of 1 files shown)


adammw:
I'd be happy with a struct too, but then you're essentially building a Location class in ZAS. Not that it's a bad thing, just then we'd have duplication with ZAM.

sandlerr:
@adammw not sure I understand how using a struct would increase duplication

danielbreves:
We'll need to make sure we do a minor bump after merging, since it's a breaking change.

danielbreves:
:+1:

lib/zendesk_apps_support/location.rb
```diff
@@ -3,16 +3,16 @@ module Location
...
+        'nav_bar' => { 'id' => 2, 'orderable' => true },
+        'ticket_sidebar' => { 'id' => 3, 'orderable' => true },
+        'new_ticket_sidebar' => { 'id' => 4, 'orderable' => true },
+        'user_sidebar' => { 'id' => 5, 'orderable' => true },
+        'organization_sidebar' => { 'id' => 6, 'orderable' => true },
+        'background' => { 'id' => 7, 'orderable' => false }
       },
       'zopim' => {
-        'chat_sidebar' => 8
+        'chat_sidebar' => { 'id' => 8, 'orderable' => false }
```
adammw: why is chat sidebar not reorderable?

adammw:
I'm still not that convinced that it belongs in ZAS if nothing in ZAS (or ZAT) uses it...

lib/zendesk_apps_support/location.rb
sandlerr: it only supports one app

adammw: the code doesn't know that. when we have multiple zopim apps we'd have to come back and update this... seems like a waste if we already have host-level filtering

adammw:
:+1:

PR #109, opened 2016-02-07
Allow having .scss file too

kobim:
Editors don't understand the SASS syntax inside .css file, this patch allows to maintain a .scss file instead of the regular .css file supplied.
`zat package` support available under zendesk/zendesk_apps_tools#115 .

### DIFF
--- a/lib/zendesk_apps_support/package.rb
+++ b/lib/zendesk_apps_support/package.rb
@@ -200,7 +200,8 @@ module ZendeskAppsSupport
 
     def app_css
       css_file = path_to('app.css')
-      File.exist?(css_file) ? File.read(css_file) : ''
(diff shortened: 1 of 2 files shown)


sandlerr:
Hey @kobim thanks for the PR and sorry about the extremely slow reply here! It makes a lot of sense to maintain the .scss file extension. Can you give a bit of feedback about your workflow? Specifically, did you use `zat server` to run your app? Browsers don't understand the scss format, did this cause any problems for you?

Without this patch, such an app would simply have no styles rather than a stylesheet with syntax errors.

sandlerr:
Actually, it turns out that these files are compiled down to normal css when using `zat server`

:+1: cc @ocke

ocke:
Totally :+1:

PR #110, opened 2016-02-10
[AF-248] Fix for setting no_template

ocke:
:v:

/cc @zendesk/vegemite

### Description

Fix for setting no_template to false instead of the array it should be.

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-248

### Risks
* Templates are not shown correctly.

### DIFF
--- a/spec/fixtures/legacy_app_no_template.js
+++ b/spec/fixtures/legacy_app_no_template.js
@@ -1,0 +1,55 @@
+with( ZendeskApps.AppScope.create() ) {
+  require.modules = {
+      "a.js": function(exports, require, module) {
+        var a = {
+  name: 'This is A'
+};
+
+module.exports = a;
+
+      },
+      "nested/b.js": function(exports, require, module) {
+        var b = {
+  name: 'This is B'
+};
+
+module.exports = b;
+
+      },
+    eom: undefined
+  };
+
+  var source = (function() {
+
+  return {
+    a: require('a.js'),
+
+    events: {
+      'app.activated':'doSomething'
+    },
+
+    doSomething: function() {
+      console.log(a.name);
+    }
+  };
+
+}());
+;
+}
+var app = ZendeskApps.defineApp(source)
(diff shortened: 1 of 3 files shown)


adammw:
As in it was mistakenly changed to false when it should have been an array or boolean? Is there a way we could add a test?

ocke:
Correct, i mistakenly changed it to false with my big zat/zas/zam restructure.

I will add a test.

ocke:
@zendesk/vegemite ?

spec/package_spec.rb
```diff
@@ -83,6 +83,13 @@
       expected = File.read('spec/fixtures/iframe_app.js')
       expect(js).to eq(expected)
     end
+
+    it 'should generate js ready for installation' do
```
adammw: The description isn't great, doesn't mention no template.

adammw:
Just realised this will be weird for multi host locations now. But let's just fix the bug first, :+1:

spec/fixtures/legacy_app_no_template.js
```diff
@@ -0,0 +1,55 @@
+with( ZendeskApps.AppScope.create() ) {
```
sandlerr: how is this file 'no template'?

spec/package_spec.rb
ocke: oops

spec/fixtures/legacy_app_no_template.js
ocke: Yeah, all I changed is the manifest which has a noTemplate option for ticket_sidebar, because that's all I'm testing for.

I agree it's a big file for only that tiny change, to you suggest a different solution?

sandlerr: sorry, didn't see that on L40

ocke:
@zendesk/vegemite :+1: ?

danielbreves:
:+1:

sandlerr:
:+1:

PR #111, opened 2016-02-12
Added backwards comp. test and fixed functionality

ocke:
:v:

/cc @zendesk/vegemite

### Description

Added backwards comp. test and fixed functionality

### References
* JIRA: 

### Risks


sandlerr:
do we actually have any apps that are defined like this?

danielbreves:
We shouldn't have.

ocke:
Yes we do, there were apps in tests of ZAF made with this.

danielbreves:
@ocke what do you mean by "apps in tests of ZAF"? If we wrote test apps in this format we should simply update them.

PR #112, opened 2016-03-31
Only try to install if the app exists

ilkkao:
This will not fix the root cause of the current RA but should limit the breakage to the bad app.

@gardhauge


ilkkao:
Just an idea we thought during the RA. Please close or ignore if there's better way to do this.

gardhauge:
:+1: from me, would have limited incidents to only affect pathfinder.

danielbreves:
:+1:

PR #119, opened 2016-05-27
AF-346 handle when location key is not supplied at all, fix some rubocop warnings

sandlerr:
:v:

### Description

We have code elsewhere that assumes `locations['zendesk'].keys` always returns a valid array of locations. However if the locations key was missing, that code snippet returns `[nil]`.

### References
https://zendesk.atlassian.net/browse/AF-346

### Risks
* [low] errors creating / updating no-location apps (eg requirements only)


xuebingli:
👍  LGTM

svizzari:
👍

PR #120, opened 2016-05-27
AF-346 dont treat no-location apps as framework v2

sandlerr:
@zendesk/vegemite @zendesk/wombat

### DIFF
--- a/lib/zendesk_apps_support/package.rb
+++ b/lib/zendesk_apps_support/package.rb
@@ -226,8 +226,7 @@ module ZendeskAppsSupport
 
     def legacy_non_iframe_app?
       iframe_urls = locations.values.flat_map(&:values)
-      return true if iframe_urls.empty?
-      @non_iframe ||= iframe_urls.any? { |l| l == LEGACY_URI_STUB }
+      iframe_urls.all? { |l| l == LEGACY_URI_STUB }
     end
 
     def templates
--- a/lib/zendesk_apps_support/package.rb
+++ b/lib/zendesk_apps_support/package.rb
@@ -225,7 +225,9 @@ module ZendeskAppsSupport
     private
 
     def legacy_non_iframe_app?
-      @non_iframe ||= locations.values.flat_map(&:values).any? { |l| l == LEGACY_URI_STUB }
+      iframe_urls = locations.values.flat_map(&:values)
+      return true if iframe_urls.empty?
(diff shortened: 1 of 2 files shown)


svizzari:
👍

lib/zendesk_apps_support/package.rb
```diff
@@ -225,7 +225,9 @@ def iframe_only?
     private
 
     def legacy_non_iframe_app?
-      @non_iframe ||= locations.values.flat_map(&:values).any? { |l| l == LEGACY_URI_STUB }
+      iframe_urls = locations.values.flat_map(&:values)
+      return true if iframe_urls.empty?
+      @non_iframe ||= iframe_urls.any? { |l| l == LEGACY_URI_STUB }
```
timbertson: If the empty case is `true`, would it make it clearer to invert the `any?` logic and remove the early `return`? e.g.:

```
is_iframe_location = -> (loc) { loc != LEGACY_URI_STUB }
urls = locations.values.flat_map(&:values)
not urls.all?(&is_iframe_location)
```

Also the memoization here is a bit haphazard - it won't short-circuit when this method returns `false`, nor when we hit the early return. And in any case, the line to calculate `iframe_urls` is probably more expensive than the part that's being short-circuited. I'd vote to ditch the memoization unless it's worth the extra code to get it right.

Out of curiosity: can we have a mix of legacy and modern locations?

sandlerr: we haven't disallowed it, so `all?` is probably more correct than `any?` in this case.

sandlerr:
@timbertson 💇 ?

svizzari:
Still 👍 , nice catch/optimisation @timbertson

lib/zendesk_apps_support/package.rb
```diff
@@ -225,7 +225,8 @@ def iframe_only?
     private
 
     def legacy_non_iframe_app?
-      @non_iframe ||= locations.values.flat_map(&:values).any? { |l| l == LEGACY_URI_STUB }
+      iframe_urls = locations.values.flat_map(&:values)
+      iframe_urls.all? { |l| l == LEGACY_URI_STUB }
```
sandlerr: all is always true for empty arrays as far as I can see

timbertson:
:+1:

PR #121, opened 2016-06-15
Add signedUrls parameter to appClass

adammw:
:v:

/cc @zendesk/vegemite

**needed for the `signedUrls` manifest parameter to be exposed to the framework**

### Description
Add signedUrls parameter to appClass

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-231

### Risks
* None, defaults to false (opt-in parameter)

### DIFF
--- a/lib/zendesk_apps_support.rb
+++ b/lib/zendesk_apps_support.rb
@@ -10,6 +10,7 @@ module ZendeskAppsSupport
   autoload :Package,                'zendesk_apps_support/package'
   autoload :Installed,              'zendesk_apps_support/installed'
   autoload :Installation,           'zendesk_apps_support/installation'
+  autoload :Finders,                'zendesk_apps_support/finders'
   autoload :AppRequirement,         'zendesk_apps_support/app_requirement'
   autoload :AppVersion,             'zendesk_apps_support/app_version'
   autoload :StylesheetCompiler,     'zendesk_apps_support/stylesheet_compiler'
--- a/lib/zendesk_apps_support/finders.rb
+++ b/lib/zendesk_apps_support/finders.rb
@@ -1,0 +1,32 @@
+module ZendeskAppsSupport
+  module Finders
+    class RecordNotFound < StandardError
+      def initialize(msg)
+        super(msg)
+      end
+    end
+
+    def find_by(arg)
+      all.find(&filter_by_arg(arg))
+    end
+
+    def find_by!(arg)
+      found = find_by(arg)
+      fail RecordNotFound.new("Unable to find #{self.name} with #{arg.inspect}") if found.nil?
+      found
+    end
+
+    def where(arg)
+      all.select(&filter_by_arg(arg))
+    end
+
+    private
+
+    def filter_by_arg(arg)
+      fail('More than one key-value pair found') if arg.size > 1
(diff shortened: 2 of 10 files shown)


lib/zendesk_apps_support.rb
```diff
@@ -10,6 +10,7 @@ module ZendeskAppsSupport
   autoload :Package,                'zendesk_apps_support/package'
   autoload :Installed,              'zendesk_apps_support/installed'
   autoload :Installation,           'zendesk_apps_support/installation'
+  autoload :Finders,                'zendesk_apps_support/finders'
```
sandlerr: curious, does this do anything given that the code always `extend ZendeskAppsSupport::Finders`?

lib/zendesk_apps_support/finders.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+
+    def where(arg)
+      all.select(&filter_by_arg(arg))
+    end
+
+    private
+
+    def filter_by_arg(arg)
+      fail('More than one key-value pair found') if arg.size > 1
```
sandlerr: I found this a little hard to follow. I think it might be easier if we explicitly state which values you can `find_by` and maybe have a better error message. I don't think this code is better than yours but maybe we can find a way to improve. I find converting hashes to array always a little odd, maybe just me.

```ruby
class Location
  can_be_found_by :code

  def self.can_be_found_by(attribute)
    @finders ||= []
    @finders << attribute
  end
  def self.finders
    @finders
  end
  def finders
    class.finders
  end
  def find_by(arg)
    raise "#{self.class} can only be queried using a single condition." if arg.keys.size != 1
    finder = arg.keys[0]
    raise "#{self.class} can only be queried by #{finders}" unless finders.include? finder
    value = arg.delete(finder)
    all.find {|l| l.public_send(finder) == value }
  end
end
```

lib/zendesk_apps_support.rb
adammw: not require a require call I assumed. since we aren't in rails and otherwise have to do the loading ourselves.

sandlerr: ah neat, so autoloading infers the namespace from the path, I thought that was a rails thing

lib/zendesk_apps_support/finders.rb
adammw: tbh, i copied this from @xuebingli's code in https://github.com/zendesk/zendesk_app_market/blob/master/app/models/product.rb

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -102,9 +102,10 @@ def invalid_location_error(package)
           errors = []
           manifest_locations = package.locations
           manifest_locations.find do |host, locations|
-            error = if !Location.hosts.include?(host)
+            product = Product.find_by(legacy_name: host)
```
danielbreves: Soon we'll have to support both, right?

danielbreves:
👍

adammw:
@danielbreves is the change in 66295e0 okay or should I add more stricter checking to ensure that developers don't mix old and new names?

timbertson:
:+1:

PR #123, opened 2016-07-05
[AF-371] framworkVersion < 2.0 error

ocke:
:v:

/cc @zendesk/vegemite

### Description
Error when using iframe apps and framework version < 2.0

### References
* JIRA: 

### Risks

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -34,6 +34,8 @@ en:
             invalid_requirements_types:
               one: 'requirements.json contains an invalid type: %{invalid_types}'
               other: 'requirements.json contains invalid types: %{invalid_types}'
+            multiple_channel_integrations: Specifying multiple channel integrations
+              in requirements.json is not supported.
             banner:
               invalid_format: Banner image must be a PNG file.
               invalid_size: Invalid banner image dimensions. Must be %{required_banner_width}x%{required_banner_height}px.
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -34,7 +34,6 @@ en:
             invalid_requirements_types:
               one: 'requirements.json contains an invalid type: %{invalid_types}'
               other: 'requirements.json contains invalid types: %{invalid_types}'
-            multiple_channel_integrations: 'Specifying multiple channel integrations in requirements.json is not supported.'
             banner:
               invalid_format: Banner image must be a PNG file.
               invalid_size: Invalid banner image dimensions. Must be %{required_banner_width}x%{required_banner_height}px.
@@ -44,7 +43,8 @@ en:
                 other: 'Missing required oauth fields in manifest: %{missing_keys}'
             missing_source: Could not find app.js
             style_in_template: "<style> tag in %{template}. Use an app.css file instead."
-            no_code_for_ifo_notemplate: Javascripts, stylesheets and templates are not allowed when an iframe URI or noTemplate is specified
+            no_code_for_ifo_notemplate: Javascripts, stylesheets, and templates are
+              not allowed when an iframe URI or noTemplate is specified
             invalid_default_locale: "%{defaultLocale} is not a valid default locale."
             missing_translation_file: 'Missing translation file for locale ''%{defaultLocale}''.
               Please read: http://developer.zendesk.com/documentation/apps/i18n.html'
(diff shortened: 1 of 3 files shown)


lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -210,6 +211,20 @@ def invalid_type_error(manifest)
         def missing_keys_validation_error(missing_keys)
           ValidationError.new('manifest_keys.missing', missing_keys: missing_keys.join(', '), count: missing_keys.length)
         end
+
+        def framework_version_iframe_only(manifest)
+          is_iframe_app = manifest['location'].any? do |host, locations|
```
danielbreves: Can we somehow reuse the code [here](https://github.com/zendesk/zendesk_apps_support/blob/master/lib/zendesk_apps_support/package.rb#L223-L232)?

config/locales/translations/zendesk_apps_support.yml
```diff
@@ -161,6 +161,10 @@ parts:
       title: "App builder job: invalid framework version"
       value: "%{target_version} is not a valid framework version. Available versions are: %{available_versions}."
   - translation:
+      key: "txt.apps.admin.error.app_build.old_version"
+      title: "App builder job: old framework version"
+      value: "%{target_version} is an old framework version. Available versions are: %{available_versions}."
```
danielbreves: "Iframe Only apps must target framework versions 2.0 or greater"?

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -210,6 +211,12 @@ def invalid_type_error(manifest)
         def missing_keys_validation_error(missing_keys)
           ValidationError.new('manifest_keys.missing', missing_keys: missing_keys.join(', '), count: missing_keys.length)
         end
+
+        def framework_version_iframe_only(package, manifest)
+          if (manifest['frameworkVersion'].nil? || package.iframe_only? && manifest['frameworkVersion'] < '2.0')
```
danielbreves: Is the `manifest['frameworkVersion'].nil?` check for tests only? Because otherwise this should never happen since this validation comes after `missing_framework_version`, right?

config/locales/en.yml
```diff
@@ -53,7 +53,7 @@ en:
               one: "%{invalid_locations} is an invalid location in %{host_name}."
               other: "%{invalid_locations} are invalid locations in %{host_name}."
             invalid_location_uri: "%{uri} is either an invalid location URI, refers
-              to a missing asset or does not use HTTPS."
+              to a missing asset, or does not use HTTPS."
```
danielbreves: Where did these translations come from?

config/locales/en.yml
```diff
@@ -34,7 +34,6 @@ en:
             invalid_requirements_types:
               one: 'requirements.json contains an invalid type: %{invalid_types}'
               other: 'requirements.json contains invalid types: %{invalid_types}'
-            multiple_channel_integrations: 'Specifying multiple channel integrations in requirements.json is not supported.'
```
danielbreves: Did you mean to delete this?

danielbreves: Looks like @timbertson [added this string](https://github.com/zendesk/zendesk_apps_support/commit/fc5493c5beaf54e457f7d5643e1b81abd9db0f51) directly here, we should copy it to zendesk_apps_support.yml instead of delete it.

ocke: That darn @timbertson, can't read:
`# This file is generated by rake i18n:standardize. Do NOT edit it directly.`
;-)

lib/zendesk_apps_support/validations/manifest.rb
ocke: nope, it runs all validations, always.

adammw: is a string check like this good enough? shouldn't we ideally be doing a string split and major version comparison to guard against stuff like 2.13.0 breaking against < '2.2'

adammw: maybe it's fine for 2.0, but still...

adammw:
:+1:

lib/zendesk_apps_support/validations/manifest.rb
danielbreves: good point

ocke: so should we fix this too then: https://github.com/zendesk/zendesk_app_framework/blob/6a7908f1d889f5988623045e53aab7ac75464923/lib/zaf_sdk_client.js#L29 ?

danielbreves: yup

ocke: ok

danielbreves:
Nice 👍

config/locales/en.yml
timbertson: Oops. Thanks for noticing (and fixing) !

PR #127, opened 2016-07-21
Default the parameter type to text instead of failing when it is empty

sandlerr:
@zendesk/vegemite 

- default parameter type to `text`
- extract a manifest object from the package
- extract a parameter object for manifest parameters
- add an error for duplicate keys in the manifest JSON
- bump ruby version used in development
- add pkg directory to .gitignore (generated by `rake install`)
- validate the fields we know must be Boolean
- deprecate calling manifest methods directly on the package
- add a validation for actually giving your app a name (was previously validated server-side only)
- treat `manifest.locations` consistently as a hash with product names, not whatever the user passed in
- error out if noTemplate is neither a boolean nor an array

### Risks

- [medium] app uploads with particular configurations in their manifest unexpectedly failing


danielbreves:
Why not fallback to text?

sandlerr:
This is still red but I'm going to refactor the way the manifest and parameters work to be a bit simpler (hopefully). We'll also be able to remove some duplication - for example the app market maintains its own, slightly different list of valid parameter types to what we use in ZAS.

sandlerr:
@danielbreves : done!
@zendesk/wombat : can I please get a review?

lib/zendesk_apps_support/manifest.rb
```diff
@@ -0,0 +1,106 @@
...
+    class << self
+      def private_attr_reader(*names)
+        private
+        attr_reader(*names)
+      end
+    end
+
+    LEGACY_URI_STUB = '_legacy'
+
+    RUBY_TO_JSON = {
```
sandlerr: I think this is a reasonably big win, although it will take slightly more work to add properties we can see what we are using at a glance. Also highlights inconsistency like `termsConditionsURL` but `signedUrls`

config/locales/en.yml
```diff
@@ -77,6 +77,9 @@ en:
             invalid_type_parameter:
               one: "%{invalid_types} is an invalid parameter type."
               other: "%{invalid_types} are invalid parameter types."
+            unacceptable_boolean: "%{field} must be either true or false without quotes."
```
timbertson: "without quotes" seems like pretty indirect advice to me. Could we just say it needs to be a boolean, and assume that app developers can tell booleans from strings?

lib/zendesk_apps_support/manifest.rb
```diff
@@ -0,0 +1,106 @@
...
+      @locations ||=
+        case original_locations
+        when Hash
+          original_locations
+        when Array
+          { 'zendesk' => NoOverrideHash[original_locations.map { |location| [ location, LEGACY_URI_STUB ] }] }
+        when String
+          { 'zendesk' => { original_locations => LEGACY_URI_STUB } }
+        # TODO: error out for numbers and Booleans
+        else # NilClass
```
timbertson: I'd prefer `when NilClass` (with an `else` which raises or explicitly ignores errors) rather than just having this be a comment.

Also is the above TODO supposed to be for now or later?

lib/zendesk_apps_support/manifest.rb
```diff
@@ -0,0 +1,106 @@
+# frozen_string_literal: true
+module ZendeskAppsSupport
+  class Manifest
+    class << self
+      def private_attr_reader(*names)
```
timbertson: I don't see this used anywhere -  is it still needed?

lib/zendesk_apps_support/manifest/no_override_hash.rb
```diff
@@ -0,0 +1,38 @@
...
+      class << self
+        def [](array)
+          uniques = Set.new
+          array.each do |key, _value|
+            if uniques.add?(key).nil?
+              # don't add the value to the error message because it might be '_legacy'
+              raise OverrideError, OverrideError.message_for(key)
+            end
+          end
+          super
```
timbertson: I love this class, but it seems a bit strange to implement the same check as `[]=` in a very different way. Could we build up our own result (instead of calling `super`) using `[]=` to do the check for us?

(it's a shame `super` doesn't already use `[]=` internally)

Not sure what to do about the desired different error message in that case though - we could catch and rethrow a less descriptive OverrideError, but to be honest that logic probably belongs in the user (manifest.rb) rather than this class anyway.

lib/zendesk_apps_support/manifest/no_override_hash.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+require 'json'
+
+module ZendeskAppsSupport
+  class Manifest
+    class OverrideError < StandardError
+      class << self
+        def message_for(key, original = nil, attempted = nil)
```
timbertson: Does this actually need to be one method which has multiple multiple modes? splitting it into `message_for_key(key)` and `message_for_pair(key, original, attempted)` would remove most of the complexity in this method, and would make it work properly when `original` or `attempted` is falsy.

lib/zendesk_apps_support/manifest.rb
```diff
@@ -0,0 +1,106 @@
...
+        no_template
+      end
+    end
+
+    def no_template_locations
+      no_template || []
+    end
+
+    def location?
+      locations != { 'zendesk' => {} }
```
timbertson: why is this defaulted differently to the others (in `initialize`)?

lib/zendesk_apps_support/manifest.rb
```diff
@@ -0,0 +1,106 @@
...
+        when Hash
+          original_locations
+        when Array
+          { 'zendesk' => NoOverrideHash[original_locations.map { |location| [ location, LEGACY_URI_STUB ] }] }
+        when String
+          { 'zendesk' => { original_locations => LEGACY_URI_STUB } }
+        # TODO: error out for numbers and Booleans
+        else # NilClass
+          { 'zendesk' => {} }
+        end
```
timbertson: Is this method the right place to canonicalize all legacy locations into current ones? I think it would nicely contain the backwards compatibility to the first code which sees this data, and would save a lot of code further up the stack from caring.

lib/zendesk_apps_support/manifest.rb
```diff
@@ -0,0 +1,106 @@
...
+    end
+
+    def iframe_only?
+      product_names = locations.keys
+      # if the app is not for Zendesk / Support, it must be iframe only
+      return true if product_names.any? { |name| !%w(zendesk support).include? name }
```
timbertson: should this logic depend on the framework version used, rather than locations? Or can you have v2 apps which aren't ifo?

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -42,19 +33,42 @@ def call(package)
 
         private
 
+        def boolean_error(manifest)
+          {
+            requirements_only: 'requirementsOnly',
```
timbertson: Can we get this from `RUBY_TO_JSON` rather than duplicating the JSON key? I know it won't actually change, but you've removed the duplication so well otherwise :)

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -173,55 +198,58 @@ def invalid_version_error(manifest, package)
...
               ValidationError.new(:old_version)
             end
           end
         end
+
+        # TODO: support the new location format in the no_template array and check the app actually runs in
+        # included locations
+        def no_template_format_error(manifest)
+          no_template = manifest.no_template
+          return if [false, true, nil].include? no_template
```
timbertson: Can we just make this `is_a?(Boolean)` since this gets defaulted to `false`?

timbertson:
Looking good! Lots of minor comments, but overall great stuff :)

I'm confused that `name` is now required though - I didn't think we took that from the manifest?

lib/zendesk_apps_support/manifest.rb
sandlerr: No, I was using it in an earlier iteration. Was on the fence whether to keep for future use or not.

lib/zendesk_apps_support/manifest.rb
sandlerr: yep, will make this simplification

lib/zendesk_apps_support/manifest.rb
sandlerr: I think that's what it does? Or do you mean changing `zendesk` to `support`?

lib/zendesk_apps_support/validations/manifest.rb
sandlerr: good point! Will try to eliminate JSON key names from everywhere except manifest.rb

lib/zendesk_apps_support/validations/manifest.rb
sandlerr: I don't think ruby has a `Boolean` class :/

lib/zendesk_apps_support/manifest.rb
timbertson: Yeah, I meant canonicalizing as in converting 'zendesk' to 'support' as early as possible. I don't think we ever need to know whether you're using a legacy name or not, so we might as well convert it ASAP and then never have to worry about it in other code.

lib/zendesk_apps_support/validations/manifest.rb
timbertson: Really Ruby? Really? 😿 
Well we can at least remove `nil` still :/

lib/zendesk_apps_support/manifest.rb
timbertson: I'd ditch it, but if others think it's worth keeping I don't really mind.

lib/zendesk_apps_support/manifest.rb
sandlerr: just because there's no `location` ivar or method. This felt like the cleanest way to do it.

lib/zendesk_apps_support/manifest.rb
sandlerr: Yep, we might care about it in validation if we ever decide to stop accepting the old names.

lib/zendesk_apps_support/manifest.rb
timbertson: Oh sorry, I actually misread this as `locations ||= { ... }`

In that case, it seems odd that having no location is equated to being exactly equal to the (empty) locations object structure, which might change if we e.g. add more products or change `zendesk` to `support`.

Could we implement this in a way that doesn't have to know anything about defaults? e.g. `locations.values.all?(&:empty)`? That way `{}`, `{'zendesk' => {}}` and `{'support' => {}}` are all location-less.

lib/zendesk_apps_support/manifest.rb
timbertson: True. Although actually it looks like `original_locations` would allow that already.

config/locales/translations/zendesk_apps_support.yml
```diff
@@ -204,3 +204,11 @@ parts:
       key: "txt.apps.admin.error.app_build.invalid_type_parameter.other"
       title: "App builder job: multiple invalid parameter types error"
       value: "%{invalid_types} are invalid parameter types."
+  - translation:
+      key: "txt.apps.admin.error.app_build.unacceptable_boolean"
+      title: "App builder job: this value needs to be either true or false, but something else was passed in"
+      value: "%{field} must be either true or false without quotes."
+  - translation:
+      key: "txt.apps.admin.error.app_build.invalid_no_template"
+      title: "App builder job: `noTemplate` needs to be either true, false or a valid array, but something else was passed in"
```
sandlerr: fixed. Should we create a new string to correct the one on line 138 as well?

sandlerr:
@zendesk/vegemite I think this should be green now, passed on my machine using phantom 1.9.8. We aren't too far off being able to use phantom 2 though, although its more strict on some of our specs - they're currently green despite throwing errors but phantom 2 seems to catch them.

sandlerr:
@timbertson I removed the requirement for `name`, that was totally wrong. Can I get another review?

config/locales/en.yml
```diff
@@ -47,8 +47,8 @@ en:
             style_in_template: "<style> tag in %{template}. Use an app.css file instead."
             no_code_for_ifo_notemplate: Javascripts, stylesheets, and templates are
               not allowed when an iframe URI or noTemplate is specified
-            invalid_default_locale: "%{defaultLocale} is not a valid default locale."
-            missing_translation_file: 'Missing translation file for locale ''%{defaultLocale}''.
+            invalid_default_locale: "%{default_locale} is not a valid default locale."
```
sandlerr: @chucknado @zendesk/i18n can someone please review the changes to this file (`config/locales/en.yml`)?

mariasica: In theory, this would need a new key. We will get validation errors, but once it is merged, I can update manually the placeholders for all current translations, is it ok @pdeuter ?

lib/zendesk_apps_support/manifest.rb
```diff
@@ -50,30 +42,31 @@ def no_template_locations
...
-      locations != { 'zendesk' => {} }
+      !locations.values.all?(&:empty?)
     end
 
     def locations
       @locations ||=
         case original_locations
         when Hash
-          original_locations
+          replace_legacy_locations! original_locations
```
timbertson: This mutates original locations, right? I guess we're not currently using `original_locations` but would it be hard to make a copy instead? Otherwise if we eventually use `original_locations` someone's going to get a surprise ;)

lib/zendesk_apps_support/manifest.rb
```diff
@@ -50,30 +42,31 @@ def no_template_locations
...
+          { 'support' => { original_locations => LEGACY_URI_STUB } }
         # TODO: error out for numbers and Booleans
         else # NilClass
-          { 'zendesk' => {} }
+          { 'support' => {} }
         end
+    rescue OverrideError => error
+      # if the error contains the word `_legacy` in the second sentence, let's
+      # only use the first one.
+      error.message = error.message[/([^.]*)\./] if error.message[/_legacy/]
```
timbertson: This code knows a lot about the error message format. Could we expose a couple more methods on the OverrideError class so we could write something like:

```
if error.values.any {|val| val =~ /_legacy/ }
  error.suppress_values!
end
```

(internally error could track `key_message` and `values_message`, then the `message` getter could just include the first when suppress_values has been set).

lib/zendesk_apps_support/manifest/no_override_hash.rb
```diff
@@ -4,33 +4,33 @@
...
+      attr_accessor :message
+      def initialize(key, original, attempted)
+        @key = key
+        @original = original
+        @attempted = attempted
+      end
+
+      def message
+        @message ||= "Duplicate reference in manifest: #{key}."\
+        " Initially set to #{original}, attempted overwrite to #{attempted}."
```
timbertson: Sorry if I've already asked this, but does this error need to be translateable?

timbertson:
Great stuff! (sorry I keep adding more work for you in my comments, but looks like it's nearly there...)

lib/zendesk_apps_support/manifest.rb
sandlerr: good point, nothing should mutate `original_locations`

lib/zendesk_apps_support/manifest/no_override_hash.rb
sandlerr: of course it does, will fix

sandlerr:
@timbertson 💇

timbertson:
💥 LGTM 👍

sandlerr:
@zendesk/vegemite can I get one more +1? I can do a walkthrough

config/locales/en.yml
sandlerr: thanks for looking @mariasica I've pushed a couple more I18n changes to this branch, it's now ready to merge with your +1

config/locales/translations/zendesk_apps_support.yml
```diff
@@ -204,3 +204,19 @@ parts:
       key: "txt.apps.admin.error.app_build.invalid_type_parameter.other"
       title: "App builder job: multiple invalid parameter types error"
       value: "%{invalid_types} are invalid parameter types."
+  - translation:
+      key: "txt.apps.admin.error.app_build.unacceptable_boolean"
+      title: "App builder job: this value needs to be either true or false, but something else was passed in"
+      value: "%{field} must be a Boolean value, got %{value}"
```
mariasica: What is loaded here? Can you give some examples? Many languages need to adapt to gender and number of the noun

config/locales/translations/zendesk_apps_support.yml
```diff
@@ -204,3 +204,19 @@ parts:
...
+  - translation:
+      key: "txt.apps.admin.error.app_build.invalid_no_template"
+      title: "App builder job: `noTemplate` needs to be true, false or a valid array, but something else was passed in"
+      value: "noTemplate must be set to true, false, or an array of valid locations."
```
mariasica: does noTemplate need to stay as is? If so, would you add this info in the title field? What about true and false? I guess they need to be in English too as they are code references? Please let me know

config/locales/translations/zendesk_apps_support.yml
```diff
@@ -204,3 +204,19 @@ parts:
...
+  - translation:
+      key: "txt.apps.admin.error.app_build.duplicate_reference"
+      title: "App builder job: a key was included twice in the supplied manifest file, but it is only allowed once"
+      value: "Duplicate reference in manifest: %{key}."
+  - translation:
+      key: "txt.apps.admin.error.app_build.duplicate_reference_values"
+      title: "This sentence follows txt.apps.admin.error.app_build.duplicate_reference. The values are included to help find the problem."
+      value: "Initially set to %{original}, attempted overwrite to %{attempted}."
```
mariasica: what would be the possible values loaded here in "orginal"?

config/locales/translations/zendesk_apps_support.yml
sandlerr: there are several field names, but we won't translate them, they will stay in English. They should all be singular. Some examples of possible values for `field` are: `requirements_only single_install signed_urls private`. The values for `value` are all supplied by the user though.

config/locales/translations/zendesk_apps_support.yml
sandlerr: We should definitely not translate `noTemplate`. We should also not translate `true` and `false` unless your team has established a convention to do so.

config/locales/translations/zendesk_apps_support.yml
sandlerr: both `original` and `attempted` are user-supplied :/

mariasica: Got it, then can we add quotes or something like Initially set to value "whatever", attempted overwrite to value "whatever"? That would avoid a lot of issues it would make sense

sandlerr: I've quoted the user-supplied values that have been added in this PR but it looks like there are some old unquoted ones. Do they need to be changed?

mariasica:
mostly, I needed to understand (so I can prevent the huge amount of questions from linguists). I will add any additional title edits on Friday. I will check past strings, and, if it is the case, just edit the title field and send for retranslation 👍

lib/zendesk_apps_support/manifest.rb
```diff
@@ -0,0 +1,113 @@
...
+        when String
+          { 'support' => { original_locations => LEGACY_URI_STUB } }
+        # TODO: error out for numbers and Booleans
+        else # NilClass
+          { 'support' => {} }
+        end
+    rescue OverrideError => error
+      # if the error contains the word `_legacy` in the second sentence, let's
+      # only use the first one.
+      if [error.original, error.attempted].any? { |val| val =~ /_legacy/ }
```
danielbreves: Couldn't we do this in the `OverrideError` class?

lib/zendesk_apps_support/package.rb
```diff
@@ -116,40 +116,41 @@ def compile_js(options)
...
-        iframe_only: iframe_only?
+        iframe_only: manifest.iframe_only?
       )
     end
 
     def manifest_json
-      @manifest ||= read_json('manifest.json')
+      @manifest_json ||= read_json('manifest.json')
+    end
+    deprecate :manifest_json, :manifest, 2016, 9
```
danielbreves: Do we need this? Why not just update all calls to manifest_json and then get rid of it?

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -4,35 +4,27 @@
...
+
+          if manifest.requirements_only?
+            check_errors(%i(ban_location ban_framework_version), errors, manifest)
           else
-            errors << missing_location_error(package)
-            errors << invalid_location_error(package)
-            errors << duplicate_location_error(manifest)
-            errors << missing_framework_version(manifest)
-            errors << invalid_version_error(manifest, package)
-            errors << framework_version_iframe_only(package, manifest)
```
danielbreves: I find this easier to read than the new way.

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -63,45 +78,57 @@ def oauth_error(manifest)
...
 
         def missing_location_error(package)
-          missing_keys_validation_error(['location']) unless package.has_location?
+          missing_keys_validation_error(['location']) unless package.manifest.location?
         end
 
         def invalid_location_error(package)
           errors = []
-          manifest_locations = package.locations
+          manifest_locations = package.manifest.original_locations
```
danielbreves: Feels a little odd to validate `original_locations` after we have already parsed it once.

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -173,53 +200,60 @@ def invalid_version_error(manifest, package)
...
-            manifest_version = Gem::Version.new (manifest['frameworkVersion'] || 0)
-            required_version = Gem::Version.new '2.0'
+        def location_framework_mismatch(manifest)
+          locations = manifest.locations
+          iframe_locations = locations.values.any? do |location_hash|
+            location_hash.values.any? { |url| url != ZendeskAppsSupport::Manifest::LEGACY_URI_STUB }
+          end
+          legacy_locations = (!iframe_locations && manifest.location?) || locations.values.any? do |location_hash|
+            location_hash.values.any? { |url| url == ZendeskAppsSupport::Manifest::LEGACY_URI_STUB }
+          end
```
danielbreves: Nitpick: could probably extract the common block here?

spec/manifest_spec.rb
```diff
@@ -0,0 +1,292 @@
...
+
+      it 'returns false when noTemplate is false' do
+        manifest_hash[:noTemplate] = false
+        expect(manifest.no_template?).to be_falsey
+      end
+    end
+    context 'when noTemplate is an array of locations' do
+      it 'returns false' do
+        manifest_hash[:noTemplate] = ['new_ticket_sidebar', 'nav_bar']
+        expect(manifest.no_template?).to be_falsey
```
danielbreves: `no_templateS` may be a *slightly* better name?

spec/validations/manifest_spec.rb
```diff
@@ -64,17 +67,17 @@ def create_package(parameter_hash)
...
 
   it 'should have an error when location is missing without requirements' do
-    allow(@package).to receive_messages(has_location?: false)
+    allow(@package.manifest).to receive(:location?) { false }
     expect(@package).to have_error 'Missing required field in manifest: location'
   end
 
   it 'should have an error when location is defined but requirements only is true' do
-    allow(@manifest).to receive_messages(read: JSON.generate(requirementsOnly: true, location: 1))
+    allow(@manifest_file).to receive_messages(read: JSON.generate(requirementsOnly: true, location: 'thicket_sidecar'))
```
danielbreves: Should we be setting up the manifest object instead of stubbing `read` here?

danielbreves:
Nice refactor! 💥

PR #128, opened 2016-08-01
Refactor a method

sandlerr:
This is for us to be able to store the long_description and installation_instructions in the app's zipfile, rather than as a separate manual installation step.

- `market_translations!` is now renamed to `runtime_translations`, no longer mutates its input and is a private method since it is only useful as an input to `compile_js`.
- `translations_for` is now public so that it can be used instead of `market_translations!` when persisting translations into a database when apps are uploaded.

### Risks

- [ medium ] translations stop working for new uploads

### DIFF
--- a/lib/zendesk_apps_support/package.rb
+++ b/lib/zendesk_apps_support/package.rb
@@ -140,7 +140,7 @@ module ZendeskAppsSupport
         asset_url_prefix: asset_url_prefix,
         app_class_name: app_class_name,
         author: author,
-        translations: translations_for(locale),
+        translations: runtime_translations(translations_for(locale)),
         framework_version: framework_version,
         templates: templates,
         modules: commonjs_modules,
@@ -183,13 +183,10 @@ module ZendeskAppsSupport
       end
     end
 
-    def market_translations!(locale)
-      result = translations[locale].fetch('app', {})
-      result.delete('name')
-      result.delete('description')
-      result.delete('long_description')
-      result.delete('installation_instructions')
-      result
+    def translations_for(locale)
+      trans = translations
+      return trans[locale] if trans[locale]
+      trans[self.manifest_json['defaultLocale']]
     end
 
     def has_location?
@@ -226,6 +223,16 @@ module ZendeskAppsSupport
 
     private
 
+
+    def runtime_translations(translations)
+      result = translations.dup
+      result.delete('name')
+      result.delete('description')
+      result.delete('long_description')
+      result.delete('installation_instructions')
+      result
+    end
+
     def legacy_non_iframe_app?
       iframe_urls = locations.values.flat_map(&:values)
       iframe_urls.all? { |l| l == LEGACY_URI_STUB }
@@ -239,12 +246,6 @@ module ZendeskAppsSupport
         memo[File.basename(file, File.extname(file))] = str
         memo
       end
-    end
-
-    def translations_for(locale)
-      trans = translations
-      return trans[locale] if trans[locale]
-      trans[self.manifest_json['defaultLocale']]
     end
 
     def translations
--- a/spec/package_spec.rb
+++ b/spec/package_spec.rb
@@ -172,20 +172,6 @@ describe ZendeskAppsSupport::Package do
     })
   end
 
-  describe 'market_translations! works as expected' do
-    it 'builds an app' do
-      expect(package.manifest_json['author']['name']).to eq('Ned Stark')
-      expect(package.send(:translations)).to eq({"en"=>{"app"=>{"description"=>"Quickly access bookmarked tickets. Syncs with the iPad app."}, "custom1"=>"The first custom thing"}})
-      expect(package.send(:market_translations!, 'en')).to eq({})
-      expect(package.send(:translations)).to eq({"en"=>{"app"=>{}, "custom1"=>"The first custom thing"}})
-    end
-
-    it 'builds an app with changed manifest' do
-      manifest['author']['name'] = 'Olaf'
-      expect(package.manifest_json['author']['name']).to eq('Olaf')
-    end
-  end
-
   describe 'Reading a manifest' do
     it 'fetches data from the manifest' do
       expect(package.manifest_json['location']).to eq('ticket_sidebar')
@@ -280,11 +266,11 @@ describe ZendeskAppsSupport::Package do
     end
   end
 
-  describe '#market_translations!' do
-    let(:translations) { { 'app' => { 'name' => 'Some App', 'description' => 'It does something.' } } }
+  describe '#runtime_translations' do
+    let(:translations) { { 'app' => { 'name' => 'Some App', 'description' => 'It does something.', 'everything_else': 'preserved' } } }
     let(:source) { build_app_source(additional_files: { "translations/en.json" => translations.to_json }) }
 
-    subject { package.market_translations!('en') }
+    subject { package.send :runtime_translations, package.translations_for('en')['app'] }
 
     it 'ignores "name"' do
       expect(subject['name']).to be nil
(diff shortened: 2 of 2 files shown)


sandlerr:
@zendesk/vegemite @zendesk/wombat @craig-day can I get some +1s?

lib/zendesk_apps_support/package.rb
```diff
@@ -183,13 +183,10 @@ def compiled_templates(app_id, asset_url_prefix)
...
 
-    def market_translations!(locale)
-      result = translations[locale].fetch('app', {})
-      result.delete('name')
-      result.delete('description')
-      result.delete('long_description')
-      result.delete('installation_instructions')
-      result
+    def translations_for(locale)
+      trans = translations
```
xuebingli: What benefit does this have besides saving a few keystrokes?

lib/zendesk_apps_support/package.rb
```diff
@@ -226,6 +223,16 @@ def iframe_only?
 
     private
 
+
+    def runtime_translations(translations)
+      result = translations.dup
```
xuebingli: Is `dup` here necessary?

lib/zendesk_apps_support/package.rb
```diff
@@ -226,6 +223,16 @@ def iframe_only?
...
+      result.delete('name')
+      result.delete('description')
+      result.delete('long_description')
+      result.delete('installation_instructions')
```
xuebingli: Use `Hash#except` or `Hash#except!`?

spec/package_spec.rb
```diff
@@ -280,11 +266,19 @@ def build_app_source_with_files(files)
...
+          'name' => 'Some App',
+          'description' => 'It does something.',
+          'everything_else' => 'preserved'
+        }
+      }
+    end
     let(:source) { build_app_source(additional_files: { "translations/en.json" => translations.to_json }) }
 
-    subject { package.market_translations!('en') }
+    subject { package.send :runtime_translations, package.translations_for('en')['app'] }
```
xuebingli: `fetch('app')` instead of `['app']`?

spec/package_spec.rb
```diff
@@ -293,6 +287,11 @@ def build_app_source_with_files(files)
     it 'ignores "description"' do
       expect(subject['description']).to be nil
     end
+
+    it 'preserves the rest' do
+      puts subject
```
xuebingli: is this for debugging?

spec/package_spec.rb
```diff
@@ -280,11 +266,19 @@ def build_app_source_with_files(files)
...
+        }
+      }
+    end
     let(:source) { build_app_source(additional_files: { "translations/en.json" => translations.to_json }) }
 
-    subject { package.market_translations!('en') }
+    subject { package.send :runtime_translations, package.translations_for('en')['app'] }
 
     it 'ignores "name"' do
       expect(subject['name']).to be nil
```
xuebingli: I reckon better to be explicit with `has_key?`

spec/package_spec.rb
```diff
@@ -293,6 +287,11 @@ def build_app_source_with_files(files)
     it 'ignores "description"' do
       expect(subject['description']).to be nil
     end
+
+    it 'preserves the rest' do
+      puts subject
+      expect(subject['everything_else']).to eq 'preserved'
```
xuebingli: Again, `fetch` may fit better here to tell us whether that key exists or not.

xuebingli:
👍 
Sorry for the nit picks. Since this is refactoring, we might as well try to do things properly :trollface:

lib/zendesk_apps_support/package.rb
sandlerr: that's a rails method

spec/package_spec.rb
sandlerr: why?

spec/package_spec.rb
sandlerr: yes

spec/package_spec.rb
sandlerr: maybe, but I think its better to debug failed assertions than KeyError raised in test

lib/zendesk_apps_support/package.rb
sandlerr: yes! Otherwise this method mutates the translations by deleting a bunch of them. IMO that is surprising and not always desired behaviour.

lib/zendesk_apps_support/package.rb
sandlerr: Better performance in the case of zat where calling `translations` reads in the files every time, other than that nothing. I didn't write this, just moved the method above `private`

spec/package_spec.rb
timbertson: why not? :trollface: 

I think you should always use `fetch`, unless you _expect_ a missing key. It makes failures happen earlier with easier to diagnose messages.

timbertson:
👍 

This will require breaking changes in users of this method, right (presumably just market)? Should we alias `market_translations!` or bump a major version?

spec/package_spec.rb
xuebingli: Shall it be removed then? Otherwise it would pollute the test output, wouldn't it?

lib/zendesk_apps_support/package.rb
xuebingli: Ah I see. I thought ActiveSupport had already been included.

lib/zendesk_apps_support/package.rb
xuebingli: No need to dup if immutability is the goal. I think you can do something like
```ruby
translations.tap do |result|
  result.delete(..)
  ...
end
```

lib/zendesk_apps_support/package.rb
xuebingli: I see. It makes sense to keep it for performance then

spec/package_spec.rb
xuebingli: Fetch gives us more info and thus is favorable in debugging. 

For example, if assertion fails and tells us it expects hash[key] to be something and gets nil, we don't know if key is not in hash or the value of key in hash is nil. 

Like Tim mentioned above, fetch is always preferred unless a missing key is expected.

lib/zendesk_apps_support/package.rb
sandlerr: doesn't seem to work

```ruby
irb(main):001:0> h = {a: 1, b: 2}
=> {:a=>1, :b=>2}
irb(main):002:0> h.tap {|inner_h| inner_h.delete(:a)}
=> {:b=>2}
irb(main):003:0> h
=> {:b=>2}
```

sandlerr:
@timbertson you want this to be strictly semver? ZAM's test suite will fail if run against this without the corresponding changes being made there.

lib/zendesk_apps_support/package.rb
xuebingli: Ah. My bad. Sorry for the wrong suggestion. Just looked at source code of Hash#except and dup was used there as well.

lib/zendesk_apps_support/package.rb
ocke: So what if I'm developing an app, running it with zat, and I'm changing my translations?

sandlerr: I don't see the problem? Your translations files will get reloaded every time. Search for `@is_cached` in this file.

timbertson:
I'd bump a major version if it's known-incompatible, can't hurt right?

If we think this gem is only used by us then we can do what we like. That still includes the option of bumping major versions on a whim though ;)

lib/zendesk_apps_support/package.rb
ocke: sorry, you're right. i didn't notice that. you didn't change anything. 👍

PR #132, opened 2016-09-07
HD-612 Strings: reverse sweep

pdeuter:
@zendesk/i18n 

### Risks
low.  English edit and title improvements.

### DIFF
--- a/config/locales/translations/zendesk_apps_support.yml
+++ b/config/locales/translations/zendesk_apps_support.yml
@@ -127,7 +127,7 @@ parts:
   - translation:
       key: "txt.apps.admin.error.app_build.missing_translation_file"
       title: "App builder job: missing translation file"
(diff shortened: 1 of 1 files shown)


mariasica:
👍  on green
Failures:
  1) translations should be up-to-date
     Failure/Error: expect(File.mtime(zendesk_version).to_i).to be <= File.mtime(standard_version).to_i, 
Failed examples:
rspec ./spec/i18n_spec.rb:11 # translations should be up-to-date

adammw:
@zendesk/i18n the failure above is a guard to try and prevent the zendesk_apps_support.yml from going out of sync of the en.yml (which is the strings that are actually used by the gem). For some reason, it occasionally passes even when the translations haven't been updated, as was the case here.

is it possible to incorporate the `rake i18n:standardize` task as part of the workflow of reverse sweeping strings for this project? that would minimize the likelihood of string updates sitting around until somebody realises that they need updating...

PR #133, opened 2016-09-26
Add icons to src.js so we don't have to guess in lotus!

ocke:
:v:

/cc @zendesk/vegemite

### Description
We supply icons from ZAS now, so host products don't have to guess and have them hardcoded!
Also, way more flexible.

### References
* JIRA: 

### Risks

### DIFF
--- a/lib/zendesk_apps_support/package.rb
+++ b/lib/zendesk_apps_support/package.rb
@@ -282,7 +282,7 @@ module ZendeskAppsSupport
     def location_icons
       {}.tap do |location_icons|
         manifest.locations.each do |host, locations_for_host|
-          next unless ['support'].include?(host)
+          next unless [Product::SUPPORT.name].include?(host)
 
           locations_for_host.keys.each do |location|
             next unless ['top_bar', 'nav_bar'].include?(location)
--- a/lib/zendesk_apps_support/package.rb
+++ b/lib/zendesk_apps_support/package.rb
@@ -131,6 +131,7 @@ module ZendeskAppsSupport
         source: source,
         app_settings: app_settings,
         asset_url_prefix: asset_url_prefix,
+        location_icons: location_icons,
         app_class_name: app_class_name,
         author: manifest.author,
         translations: runtime_translations(translations_for(locale)),
@@ -276,6 +277,30 @@ module ZendeskAppsSupport
 
     def app_js
       read_file('app.js')
+    end
+
+    def location_icons
+      {}.tap do |location_icons|
+        manifest.locations.each do |host, locations_for_host|
+          next unless ['support'].include?(host)
+
+          locations_for_host.keys.each do |location|
+            next unless ['top_bar', 'nav_bar'].include?(location)
+            location_icons[host] ||= {}
+            location_icons[host][location] = if (has_file?("assets/icon_#{location}.svg"))
+              { 'svg' => "icon_#{location}.svg" }
+            elsif (has_file?("assets/icon_#{location}_inactive.png"))
+              {
+                'inactive' => "icon_#{location}_inactive.png",
+                'active' => has_file?("assets/icon_#{location}_active.png") ? "icon_#{location}_active.png" : "icon_#{location}_inactive.png",
+                'hover' => has_file?("assets/icon_#{location}_hover.png") ? "icon_#{location}_hover.png" : "icon_#{location}_inactive.png"
+              }
+            else
+              {}
+            end
+          end
+        end
+      end
     end
 
     def commonjs_modules
--- a/lib/zendesk_apps_support/package.rb
+++ b/lib/zendesk_apps_support/package.rb
@@ -15,6 +15,8 @@ module ZendeskAppsSupport
     DEFAULT_SCSS   = File.read(File.expand_path('../assets/default_styles.scss', __FILE__))
     SRC_TEMPLATE   = Erubis::Eruby.new(File.read(File.expand_path('../assets/src.js.erb', __FILE__)))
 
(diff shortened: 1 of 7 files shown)


sandlerr:
add some tests please

lib/zendesk_apps_support/package.rb
```diff
@@ -278,6 +279,28 @@ def app_js
...
 
+    def location_icons
+      location_icons = {}
+      manifest.locations.each do |host, locations|
+        locations.keys.each do |location|
+          location_icons[host] = {} unless location_icons[host]
+
+          if (has_file?("assets/icon_#{location}.svg"))
+            location_icons[host][location] = { svg: "icon_#{location}.svg" }
+
```
adammw: nit: extra whitespace

ocke: ?

adammw: there's more whitespace on this branch of the condition than the other

lib/zendesk_apps_support/package.rb
```diff
@@ -278,6 +279,28 @@ def app_js
       read_file('app.js')
     end
 
+    def location_icons
+      location_icons = {}
+      manifest.locations.each do |host, locations|
+        locations.keys.each do |location|
```
adammw: Wasn't there a different/better way to get locations now? (maybe not, i can't remember)

danielbreves:
@ocke do you have a matching PR where this is used?

lib/zendesk_apps_support/package.rb
```diff
@@ -14,6 +14,7 @@ class Package
     DEFAULT_LAYOUT = Erubis::Eruby.new(File.read(File.expand_path('../assets/default_template.html.erb', __FILE__)))
     DEFAULT_SCSS   = File.read(File.expand_path('../assets/default_styles.scss', __FILE__))
     SRC_TEMPLATE   = Erubis::Eruby.new(File.read(File.expand_path('../assets/src.js.erb', __FILE__)))
+    DEFAULT_IMAGE_CDN = 'https://cdn.zendesk.com/assets'
```
ocke: we need to change this to the correct one.

ocke:
https://github.com/zendesk/zendesk_console/pull/11982

lib/zendesk_apps_support/package.rb
```diff
@@ -278,6 +279,30 @@ def app_js
       read_file('app.js')
     end
 
+    def location_icons
+      {}.tap do |location_icons|
+        manifest.locations.each do |host, locations_for_host|
+          next unless ['support'].include?(host)
```
adammw: why do we only do this for support? and if we do still need this guard, could use `ZendeskAppsSupport::Product::SUPPORT.name`?

lib/zendesk_apps_support/package.rb
```diff
@@ -278,6 +279,30 @@ def app_js
       read_file('app.js')
     end
 
+    def location_icons
+      {}.tap do |location_icons|
+        manifest.locations.each do |host, locations_for_host|
+          next unless [Product::SUPPORT.name].include?(host)
+
+          locations_for_host.keys.each do |location|
+            next unless ['top_bar', 'nav_bar'].include?(location)
```
svizzari: Worth extracting the locations into a constant?

lib/zendesk_apps_support/package.rb
```diff
@@ -278,6 +279,30 @@ def app_js
...
+            location_icons[host] ||= {}
+            location_icons[host][location] = if (has_file?("assets/icon_#{location}.svg"))
+              { 'svg' => "icon_#{location}.svg" }
+            elsif (has_file?("assets/icon_#{location}_inactive.png"))
+              {
+                'inactive' => "icon_#{location}_inactive.png",
```
svizzari: Like above, worth extracting a constant for `ICON_STATES` and iterating over that rather than the duplication?

spec/package_spec.rb
```diff
@@ -350,4 +350,98 @@ def build_app_source_with_files(files)
...
+              'svg' => "icon_nav_bar.svg"
+            }
+          }
+        })
+      end
+    end
+
+    context 'when it has three pngs' do
+      it 'returns correct location_icons hash' do
+        allow(package).to receive(:has_file?) { |file| file != 'assets/icon_top_bar.svg' && file != 'assets/icon_nav_bar.svg'}
```
sandlerr: does this return true for the png files?

ocke: yes exactly.

lib/zendesk_apps_support/package.rb
danielbreves: Could we extract `"icon_#{location}_inactive.png"` to a variable, like `default_png_icon`?

sandlerr:
@danielbreves hang on, you just gave this PR a :+1: but asked about v1 apps defining `locationIcons` on the lotus PR. Shouldn't we solve that problem here, not in lotus?

danielbreves:
@sandlerr not necessarily, whichever name we pick could cause conflicts. Perhaps we should change the Framework to not blindly merge these properties with third-party object, and instead store them separately under a stricter namespace?

lib/zendesk_apps_support/package.rb
ocke: I gave it a try and we end up with more code and IMHO no added readability.

svizzari: This line and the two below are very similar, the only variation is the state name and they all fall back to inactive.  Couldn't you store the states in an array, iterate over them and replace the state name part?

PR #134, opened 2016-10-17
Add frameworkOptions to manifest

adammw:
:v:

/cc @zendesk/vegemite

### Description
Add `frameworkOptions` to manifest, with the intention that we can add more options to it and they get passed through directly to ZAF without further ZAS changes. `noTemplate`, `singleInstall` and `signedUrls` will be supported on the root for backwards compatibility.

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-504

### Risks
* low: locations, signed urls, or single install parameters go wonky

### DIFF
--- a/lib/zendesk_apps_support/manifest.rb
+++ b/lib/zendesk_apps_support/manifest.rb
@@ -8,6 +8,7 @@ module ZendeskAppsSupport
       marketing_only: 'marketingOnly',
       version: 'version',
       author: 'author',
+      framework_options: 'frameworkOptions',
       framework_version: 'frameworkVersion',
       single_install: 'singleInstall',
       signed_urls: 'signedUrls',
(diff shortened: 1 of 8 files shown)


lib/zendesk_apps_support/manifest.rb
```diff
@@ -101,6 +103,15 @@ def set_locations_and_hosts
         end
     end
 
+    def set_framework_options
+      @framework_options ||= {
+        noTemplate: no_template_locations,
+        singleInstall: single_install?,
```
danielbreves: I'm not sure `singleInstall` is a framework option, seems more like a marketplace option.

adammw: hmm, well maybe this naming scheme doesn't work that well then...
https://github.com/zendesk/zendesk_app_framework/blob/01f77fca6d3b00684d47706205d23bd7c8beabb9/lib/app/base.js#L123

danielbreves: Oh, right, nvm then.

adammw:
Closing in favour of #135

PR #135, opened 2016-10-17
Add experiments option to manifest available to ZAF

adammw:
:v:

/cc @zendesk/vegemite

### Description
Add experiments option to manifest available to ZAF

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-504

### Risks
* low: breaks app serialization

### DIFF
--- a/lib/zendesk_apps_support/manifest.rb
+++ b/lib/zendesk_apps_support/manifest.rb
@@ -67,6 +67,10 @@ module ZendeskAppsSupport
       end
     end
 
+    def enabled_experiments
+      (experiments || {}).select { |k, v| v }.keys
+    end
+
     def initialize(manifest_text)
       m = parse_json(manifest_text)
       RUBY_TO_JSON.each do |ruby, json|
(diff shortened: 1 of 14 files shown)


danielbreves:
Can we add a marketplace validation to ensure we only allow uploading apps with whitelisted experiment keys?

adammw:
💇  added validation to a whitelist, that is empty for now.

lib/zendesk_apps_support/validations/marketplace.rb
```diff
@@ -14,6 +16,13 @@ def no_symlinks(path)
           end
           nil
         end
+
+        def no_experiments(manifest)
+          non_public_experiments = manifest.enabled_experiments - WHITELISTED_EXPERIMENTS
```
sandlerr: that's user supplied input, not necessarily `non_public` but more like `invalid`?

lib/zendesk_apps_support/manifest.rb
```diff
@@ -66,6 +67,10 @@ def parameters
       end
     end
 
+    def enabled_experiments
+      (experiments || {}).select { |k, v| v }.keys
```
sandlerr: sorry, misunderstood.

lib/zendesk_apps_support/validations/marketplace.rb
adammw: will change the name

PR #137, opened 2016-10-24
AF-526 modals don't work with signed URLs

sandlerr:
### Description

Add the modal location so that we can store the URL to redirect for signing.
cc @zendesk/vegemite 

### References

https://zendesk.atlassian.net/browse/AF-526

### Risks
- [low] failing app uploads / updates


lib/zendesk_apps_support/location.rb
```diff
@@ -3,8 +3,12 @@ class Location
     extend ZendeskAppsSupport::Finders
     attr_reader :id, :name, :orderable, :product_code
 
+    IDS = Set.new
```
danielbreves: This isn't really a constant, right? Can we make it lower case, maybe `unique_ids`?

sandlerr: `add` is only called from within the declaration of `LOCATIONS_AVAILABLE`

danielbreves: Right now it may be, but what prevents a new location from being created from anywhere else, causing another call to `add`?

sandlerr: updated

PR #140, opened 2016-11-08
[AF-486] standalone_chat location

ocke:
:v:

/cc @zendesk/vegemite

### Description
Allow lotus box as a location in the form of 'standalone_chat'

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-486

### Risks

### DIFF
--- a/lib/zendesk_apps_support/location.rb
+++ b/lib/zendesk_apps_support/location.rb
@@ -39,7 +39,7 @@ module ZendeskAppsSupport
       Location.new(id: 11, orderable: false, name: 'top_bar', product_code: Product::STANDALONE_CHAT.code),
       Location.new(id: 12, orderable: false, name: 'nav_bar', product_code: Product::STANDALONE_CHAT.code),
       Location.new(id: 13, orderable: false, name: 'system_top_bar', product_code: Product::SUPPORT.code),
-      Location.new(id: 11, orderable: false, name: 'system_top_bar', product_code: Product::STANDALONE_CHAT.code)
+      Location.new(id: 14, orderable: false, name: 'system_top_bar', product_code: Product::STANDALONE_CHAT.code)
     ].freeze
   end
 end
--- a/lib/zendesk_apps_support/location.rb
+++ b/lib/zendesk_apps_support/location.rb
@@ -37,7 +37,9 @@ module ZendeskAppsSupport
       Location.new(id: 9, orderable: false, name: 'modal', product_code: Product::SUPPORT.code),
       Location.new(id: 10, orderable: false, name: 'ticket_editor', product_code: Product::SUPPORT.code),
       Location.new(id: 11, orderable: false, name: 'top_bar', product_code: Product::STANDALONE_CHAT.code),
-      Location.new(id: 12, orderable: false, name: 'nav_bar', product_code: Product::STANDALONE_CHAT.code)
+      Location.new(id: 12, orderable: false, name: 'nav_bar', product_code: Product::STANDALONE_CHAT.code),
+      Location.new(id: 13, orderable: false, name: 'system_top_bar', product_code: Product::SUPPORT.code),
+      Location.new(id: 11, orderable: false, name: 'system_top_bar', product_code: Product::STANDALONE_CHAT.code)
     ].freeze
   end
 end
--- a/lib/zendesk_apps_support/location.rb
+++ b/lib/zendesk_apps_support/location.rb
@@ -35,7 +35,9 @@ module ZendeskAppsSupport
       Location.new(id: 7, orderable: false, name: 'background', product_code: Product::SUPPORT.code),
       Location.new(id: 8, orderable: true, name: 'chat_sidebar', product_code: Product::CHAT.code),
       Location.new(id: 9, orderable: false, name: 'modal', product_code: Product::SUPPORT.code),
-      Location.new(id: 10, orderable: false, name: 'ticket_editor', product_code: Product::SUPPORT.code)
+      Location.new(id: 10, orderable: false, name: 'ticket_editor', product_code: Product::SUPPORT.code),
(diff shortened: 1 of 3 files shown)


lib/zendesk_apps_support/location.rb
```diff
@@ -35,7 +35,9 @@ def self.all
       Location.new(id: 7, orderable: false, name: 'background', product_code: Product::SUPPORT.code),
       Location.new(id: 8, orderable: true, name: 'chat_sidebar', product_code: Product::CHAT.code),
       Location.new(id: 9, orderable: false, name: 'modal', product_code: Product::SUPPORT.code),
-      Location.new(id: 10, orderable: false, name: 'ticket_editor', product_code: Product::SUPPORT.code)
+      Location.new(id: 10, orderable: false, name: 'ticket_editor', product_code: Product::SUPPORT.code),
+      Location.new(id: 11, orderable: false, name: 'top_bar', product_code: Product::STANDALONE_CHAT.code),
```
danielbreves: What about `system_top_bar`?

lib/zendesk_apps_support/package.rb
```diff
@@ -190,7 +190,7 @@ def has_location?
     deprecate :has_location?, 'manifest.location?', 2016, 9
 
     def has_file?(path)
-      File.exist?(path_to(path))
+      File.file?(path_to(path))
```
sandlerr: did we ever use this to check if there's an assets folder or anything like that?

lib/zendesk_apps_support/location.rb
```diff
@@ -35,7 +35,11 @@ def self.all
...
+      Location.new(id: 12, orderable: false, name: 'nav_bar', product_code: Product::STANDALONE_CHAT.code),
+      Location.new(id: 13, orderable: false, name: 'system_top_bar', product_code: Product::SUPPORT.code),
+      Location.new(id: 11, orderable: false, name: 'system_top_bar', product_code: Product::STANDALONE_CHAT.code)
```
sandlerr: duplicate ID, that's why the tests are red :D
does the system_top_bar location have to exist in both products?

lib/zendesk_apps_support/package.rb
ocke: no i checked. the function is only used for files.

lib/zendesk_apps_support/location.rb
ocke: oops

ocke: well, i guess so, because they will use the same apis that will not be available in top_bar.

lib/zendesk_apps_support/package.rb
ocke: it's also used for validating location url's:
```
  location: { support: { top_bar: "my_page.html" } }
```
can that be a path to a folder?

lib/zendesk_apps_support/location.rb
sandlerr: I don't think this one needs to exist - the only app in Product::STANDALONE_CHAT will use system_top_bar right?

lib/zendesk_apps_support/package.rb
sandlerr: No, that wouldn't work (and even if it did, we never documented it as supported).

ocke: so that means my change is :OK: :-)

lib/zendesk_apps_support/location.rb
ocke: actually, yes you're right.

sandlerr:
standalone_chat is a confusing name though, it's not for apps that run in chat, it's for running the chat product as an app.. let's try and come up with a better name.

adammw:
@sandlerr the standalone name comes from the product names internal in lotus box, while it's not that clear, i would think keeping the 1:1 mapping would be nicer than describing it by more names at this point (considering it's supposed to be temporary anyway)

lib/zendesk_apps_support/product.rb
```diff
@@ -16,10 +16,12 @@ def self.all
     # The product codes below match the values in the database, do not change them!
     PRODUCTS_AVAILABLE = [
       Product.new(code: 1, name: 'support', legacy_name: 'zendesk'),
-      Product.new(code: 2, name: 'chat', legacy_name: 'zopim')
+      Product.new(code: 2, name: 'chat', legacy_name: 'zopim'),
+      Product.new(code: 3, name: 'standalone_chat', legacy_name: 'lotus_box')
```
adammw: can we leave legacy_name nil (assuming it has never been used as `lotus_box`)?

lib/zendesk_apps_support/package.rb
```diff
@@ -284,7 +284,7 @@ def app_js
     def location_icons
       {}.tap do |location_icons|
         manifest.locations.each do |host, locations_for_host|
-          next unless [Product::SUPPORT.name].include?(host)
+          next unless [Product::SUPPORT.name, Product::STANDALONE_CHAT.name].include?(host)
```
ocke: yeah.

sandlerr:
let's skip [this](https://github.com/zendesk/zendesk_apps_support/blob/344b410e0fdb5c128860f66832666ce3d76e2eba/lib/zendesk_apps_support/manifest.rb#L114-L115) when the legacy_key is nil

sandlerr:
should also make a PR to handle it in https://github.com/zendesk/zendesk_app_market/blob/d0b83d5190878ae33533f326f6905f2924d594cf/app/controllers/api/v2/locations_controller.rb#L52 and https://github.com/zendesk/zendesk_app_market/blob/4f4b478f3ecd5508eeb27e0ed65980ce60e76255/lib/app_market/app_package/builder.rb#L156

ocke:
@sandlerr I actually changed the nil back to 'lotus_box' because I don't want to start patching all these products, because the code that uses the object should support nil.

PR #141, opened 2016-11-13
Add name to manifest

ocke:
:v:

/cc @zendesk/vegemite

### Description
Make name a thing

### References
* JIRA: 

### Risks
- [high] roman unpleased

### DIFF
--- a/lib/zendesk_apps_support/manifest.rb
+++ b/lib/zendesk_apps_support/manifest.rb
@@ -8,6 +8,7 @@ module ZendeskAppsSupport
       marketing_only: 'marketingOnly',
       version: 'version',
       author: 'author',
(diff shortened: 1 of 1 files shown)


sandlerr:
won't this cause confusion though because it is only used under zat?

ocke:
@sandlerr I added a comment to say it's only used in ZAT. Currently. Because I would like to change that.

sandlerr:
a comment is good, but most people don't look at the code, they only run it. What about emitting a warning on the console?

sandlerr:
^ as a separate PR is fine

ocke:
I added a message here: https://github.com/zendesk/zendesk_apps_tools/pull/158

PR #142, opened 2016-11-21
[AF-527] Support location objects in manifest

adammw:
:v: 

/cc @zendesk/vegemite

⛔️ **depends on https://github.com/zendesk/zendesk_app_framework/pull/657** ⛔️ 

### Description
Support location objects in manifest
e.g. 
```
{
  "location": {
    "support": {
      "background": {
        "url": "assets/iframe.html"
      },
      "nav_bar": {
        "url": "assets/iframe.html",
        "autoHide": true
      },
      "ticket_sidebar": {
        "url": "assets/iframe.html",
        "autoLoad": false
      },
      "new_ticket_sidebar": {
        "autoLoad": false
      }
    }
}
```

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-527

### Risks
* medium: generates source with a different location format

### DIFF
--- a/lib/zendesk_apps_support/validations/manifest.rb
+++ b/lib/zendesk_apps_support/validations/manifest.rb
@@ -158,11 +158,10 @@ module ZendeskAppsSupport
 
             locations.each do |location_key, location|
               url = location['url']
-              no_iframe = location.fetch('noIframe', false)
+              auto_load = location.fetch('autoLoad', true)
               if url && !url.empty?
-                errors << ValidationError.new(:cant_use_uri_with_no_iframe, location: location_key) if no_iframe
                 errors << invalid_location_uri_error(package, location['url'])
-              elsif !no_iframe
+              elsif auto_load
                 errors << ValidationError.new(:blank_location_uri, location: location_key)
               end
             end
--- a/lib/zendesk_apps_support/validations/manifest.rb
+++ b/lib/zendesk_apps_support/validations/manifest.rb
@@ -156,8 +156,8 @@ module ZendeskAppsSupport
                                   count: invalid_locations.length)
             end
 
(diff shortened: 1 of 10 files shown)


lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -156,8 +156,15 @@ def invalid_location_error(package)
                                   count: invalid_locations.length)
             end
 
-            locations.values.each do |path|
-              errors << invalid_location_uri_error(package, path)
+            locations.each do |location_key, location|
+              url = location['url']
+              no_iframe = location.fetch('noIframe', false)
+              if url && !url.empty?
```
sandlerr: why the empty check?

sandlerr:
now that this is an object rather than a string, is it a good time to stop using the `_legacy` hack?

adammw:
😅  i'd love to (i.e. some sort of "legacy": true thing), but we're going to need to support _legacy for some time anyway since it gets serialised into the app source. one step at a time :)

sandlerr:
right, thought now would be a good time to introduce a new way

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -156,8 +156,14 @@ def invalid_location_error(package)
                                   count: invalid_locations.length)
             end
 
-            locations.values.each do |path|
-              errors << invalid_location_uri_error(package, path)
+            locations.each do |location_key, location|
+              url = location['url']
+              auto_load = location.fetch('autoLoad', true)
+              if url && !url.empty?
```
sandlerr: why the empty check? we should already be checking valid url somewhere?

adammw:
Closing as #144 has been merged

PR #143, opened 2016-11-22
[AF-528] Prevent noTemplate without locations from being used in v2 apps	

adammw:
:v:

/cc @zendesk/vegemite

### Description
Prevent noTemplate without locations from being used in v2 apps.
Translation references `autoLoad` property which is only really allowed after https://github.com/zendesk/zendesk_apps_support/pull/142

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-528

### Risks
* low: prevents noTemplate from validating in certain scenarios


config/locales/en.yml
```diff
@@ -9,6 +9,8 @@ en:
             jshint:
               one: 'JSHint error in %{file}: %{errors}'
               other: 'JSHint errors in %{file}: %{errors}'
+            no_template_deprecated_in_v2: noTemplate option is deprecated and cannot
```
sandlerr: I think this reads better without "option". Maybe

> noTemplate is deprecated and cannot be used with frameworkVersion 2 or above. Set the noIframe option for specific locations instead <interpolate link to doc so we don't have to re-translate>.

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -70,6 +72,11 @@ def check_errors(error_types, collector, *checked_objects)
           end
         end
 
+        def ban_no_template(manifest)
+          no_template_migration_link = 'https://developer.zendesk.com/apps/docs/apps-v2/changelog'
```
danielbreves: This is not the final link, right?

danielbreves:
Actually, thinking about this again, it might make sense to [change the behaviour of Nav Bar/Top Bar noTemplate apps](https://docs.google.com/document/d/1JtkdYq7KzD7GisdnzX2IS7KGgX9b6-KponMUQrRXMws/edit?disco=AAAAA4IRE3w). In which case we may want to keep supporting noTemplate for legacy apps migrating to v2.

adammw:
@danielbreves mind you this is preventing only `noTemplate: true` not `noTemplate: ['ticket_sidebar']`, and validations don't prevent it from working in zat.

lib/zendesk_apps_support/validations/manifest.rb
adammw: should I reference the manifest#location instead you think?

danielbreves: :+1:

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -71,6 +73,11 @@ def check_errors(error_types, collector, *checked_objects)
           end
         end
 
+        def ban_no_template(manifest)
+          no_template_migration_link = 'https://developer.zendesk.com/apps/docs/apps-v2/manifest#location'
+          ValidationError.new(:no_template_deprecated_in_v2, link: no_template_migration_link) if manifest.no_template?
```
sandlerr: shouldn't we check `manifest.no_template_locations.any?` here too?

adammw:
:haircut_man: this now prevents both noTemplate: true and noTemplate: [array of locations]

danielbreves:
@dpawluk, this may surprise some developers, though I don't think it's a breaking change per se, since we never documented `noTemplate` in v2.

danielbreves:
cc @zendesk/zdbreaking

dpawluk:
@danielbreves totally good to go with this one, thanks for notification 👍

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -71,6 +73,13 @@ def check_errors(error_types, collector, *checked_objects)
           end
         end
 
+        def ban_no_template(manifest)
+          no_template_migration_link = 'https://developer.zendesk.com/apps/docs/apps-v2/manifest#location'
```
sandlerr: nit: this can go inside the if statement

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -71,6 +73,13 @@ def check_errors(error_types, collector, *checked_objects)
...
+          if manifest.no_template? || !manifest.no_template_locations.empty?
```
sandlerr: why not `manifest.no_template_locations.any?` ?

spec/validations/manifest_spec.rb
```diff
@@ -82,12 +82,17 @@ def create_package(parameter_hash)
     expect(@package).not_to have_error 'Missing required field in manifest: location'
   end
 
+  it 'should have an error when app is iframe only but specifies noTemplate: true' do
+    @manifest_hash = { noTemplate: true, frameworkVersion: '2.0.0' }
```
sandlerr: 🙏  1 more test for `noTemplate` array?

adammw: 608c52a

sandlerr:
`spec/validations/manifest_spec.rb:90:1: C: Extra blank line detected.`

dpawluk:
I know you all said this might still work in ZAT. Is that on the docket to update at some point in the future too?

sandlerr:
@dpawluk I can release ZAT today, but in the meantime you can test this in zat after running `gem update zendesk_apps_support`

adammw:
@dpawluk what I was referring to there was that `zat serve` doesn't enforce the validations, you can serve an app that doesn't validate, however you won't be able to zip or upload it...

dpawluk:
@adammw cool, I just wanted to make sure validation in ZAT was as close to uploading validation as possible. Thanks!

PR #144, opened 2016-12-01
AF-527 location options

sandlerr:
### Description

- Converts the v2 location URL string to an object with a URL property and other properties:
- Adds support for `autoLoad` and `autoHide` props for each v2 location
- Create a `LocationOptions` class representing... the options for a location
- Adds rubocop configuration and a travis job (allowed failure for now)
- Remove validation and error strings that can't happen anymore
- Delete deprecated location methods on Package class

### References

https://zendesk.atlassian.net/browse/AF-527

### Risks

- [medium] bugs in new uploads not showing in the right location, or not being hidden on load when they should be


adammw:
For reference, this is an alternative approach building off https://github.com/zendesk/zendesk_apps_support/pull/142

lib/zendesk_apps_support/i18n.rb
```diff
@@ -9,12 +9,16 @@ def set_locale(locale)
         i18n.locale = locale
       end
 
+      def set_load_path
+        require 'i18n'
+        ::I18n.load_path += locale_files
```
sandlerr: actually, this should just call `i18n`, otherwise we might double-append.

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -139,31 +138,32 @@ def default_locale_error(manifest, package)
...
-          manifest_locations.find do |host, locations|
-            product = Product.find_by(name: host)
-            stub = ZendeskAppsSupport::Manifest::LEGACY_URI_STUB
-            locations_allowed = Location.where(product_code: product.code).map(&:name).push(stub)
-            if (invalid_locations = locations.keys - locations_allowed).any?
-              errors << ValidationError.new(:invalid_location,
-                                  invalid_locations: invalid_locations.join(', '),
-                                  host_name: host,
-                                  count: invalid_locations.length)
+          package.manifest.location_options.each do |location_options|
```
danielbreves: A little strange that the item is the same name as the collection, maybe we could use singular `location_option`?

spec/spec_helper.rb
```diff
@@ -2,6 +2,7 @@
 require 'i18n'
 
 I18n.enforce_available_locales = false
+ZendeskAppsSupport::I18n.set_load_path
```
danielbreves: Why did you need to do this?

lib/zendesk_apps_support/finders.rb
```diff
@@ -23,10 +23,12 @@ def where(arg)
     private
 
     def filter_by_arg(arg)
-      fail('More than one key-value pair found') if arg.size > 1
-      attribute, value = arg.to_a.first
-      value = value.to_s if value.is_a? Symbol
-      ->(product) { product.public_send(attribute) == value }
+      lambda do |product|
```
danielbreves: Kinda strange that we name the argument `product` here, when everything else in the module has a generic name.

danielbreves:
This seems like medium risk rather than low, can we make sure to QA this in ZAM staging before it hits production?

cc @alieu

lib/zendesk_apps_support/package.rb
```diff
@@ -282,23 +264,25 @@ def app_js
...
-            elsif (has_file?("assets/icon_#{location}_inactive.png"))
-              {
-                'inactive' => "icon_#{location}_inactive.png",
-                'active' => has_file?("assets/icon_#{location}_active.png") ? "icon_#{location}_active.png" : "icon_#{location}_inactive.png",
-                'hover' => has_file?("assets/icon_#{location}_hover.png") ? "icon_#{location}_hover.png" : "icon_#{location}_inactive.png"
-              }
-            else
-              {}
-            end
+      Hash.new { |h, k| h[k] = {} }.tap do |location_icons|
```
sandlerr: not quite - here's the difference in behaviour:

```ruby
irb(main):001:0>  no_block = Hash.new({})
=> {}
irb(main):002:0> no_block[:a][:a] = 'hello'
=> "hello"
irb(main):003:0> no_block[:b][:b] = 'world'
=> "world"
irb(main):004:0> no_block[:c]
=> {:a=>"hello", :b=>"world"}
irb(main):005:0> no_block.inspect
=> "{}"
irb(main):006:0> with_block = Hash.new {|h, k| h[k] = {} }
=> {}
irb(main):007:0> with_block[:a][:a] = 'hello'
=> "hello"
irb(main):008:0> with_block[:b][:b] = 'world'
=> "world"
irb(main):009:0> with_block[:c]
=> {}
irb(main):010:0> with_block.inspect
=> "{:a=>{:a=>\"hello\"}, :b=>{:b=>\"world\"}, :c=>{}}"
```

lib/zendesk_apps_support/validations/manifest.rb
sandlerr: thinking of renaming `LocationOptions` to `LocationOptionGroup`, what do you think?

lib/zendesk_apps_support/package.rb
sandlerr: that is, the `no_block` option returns the *same* hash every time, and doesn't actually assign it to that key on the parent hash.

spec/spec_helper.rb
danielbreves: I'm probably missing something but It seems like every public method in `I18n` calls `i18n`?

lib/zendesk_apps_support/validations/manifest.rb
danielbreves: Not sure `group` is the right word, it might make it more confusing...

lib/zendesk_apps_support/package.rb
danielbreves: Ah I see, misread the docs.

danielbreves: It did make me have to go look for docs though, wondering if it might be better to explicitly assign the value each time?

spec/spec_helper.rb
sandlerr: yeah, must be code somewhere talking to `::I18n` rather than `ZAS::I18n`

lib/zendesk_apps_support/validations/manifest.rb
sandlerr: what's a better name? `location_option` isn't right because it isn't one option, it's at least four corresponding to a particular location.

lib/zendesk_apps_support/finders.rb
sandlerr: renamed

sandlerr:
risk level has been bumped

lib/zendesk_apps_support/validations/manifest.rb
adammw: maybe we rename `location_option` to `locations` (and `locations` to `location_hash`) ?

config/locales/translations/zendesk_apps_support.yml
```diff
@@ -145,17 +145,23 @@ parts:
       title: "App builder job: invalid locations"
       value: "%{invalid_locations} are invalid locations in %{host_name}."
   - translation:
+      key: "txt.apps.admin.error.app_build.blank_location_uri"
+      title: "App builder job: invalid URI for an iframe in the manifest"
+      value: "%{location} location does not specify a URI."
```
mariasica: hello! could you give me an example of complete sentence here? what exactly is loaded in the placeholder, a path?

sandlerr: `location` will be a value like `ticket_sidebar`, `top_bar`, etc

> ticket_sidebar location does not specify a URI.

PR #148, opened 2016-12-13
dont double-declare constants

sandlerr:
### Description

Cleanup of some variables that are also allowed by telling JSHint to validate for the browser environment.

cc @zendesk/vegemite

### Risks

- [medium] unexpected JSHint errors in v1 apps


adammw:
Not sure I understand... is this because we provide our own versions of those globals?

sandlerr:
they've just been left there, originally included before we set `browser: true`

PR #149, opened 2016-12-13
swap sass for sassc

sandlerr:
### Description

Swap the ruby sass compiler for libsass (ruby bindings provided by sassc). The different security policies are a good reason to investigate this change: https://github.com/sass/sass/issues/1011#issuecomment-30037620 and https://github.com/sass/libsass/blob/master/SECURITY.md

cc @zendesk/vegemite @zendesk/zdbreaking 

### Risks

- [medium] sass that used to compile might now product different results

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -7,6 +7,7 @@ PATH
       image_size
       jshintrb (~> 0.3.0)
       json
+      sass
       sassc
 
 GEM
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -7,7 +7,7 @@ PATH
       image_size
       jshintrb (~> 0.3.0)
       json
-      sass
+      sassc
 
 GEM
   remote: https://rubygems.org/
@@ -20,6 +20,7 @@ GEM
     execjs (2.7.0)
     faker (1.6.6)
       i18n (~> 0.5)
+    ffi (1.9.14)
     i18n (0.7.0)
     image_size (1.5.0)
     jshintrb (0.3.0)
@@ -32,7 +33,7 @@ GEM
       ast (~> 2.2)
     powerpack (0.1.1)
     rainbow (2.1.0)
-    rake (11.3.0)
+    rake (12.0.0)
     rspec (3.4.0)
       rspec-core (~> 3.4.0)
       rspec-expectations (~> 3.4.0)
@@ -54,6 +55,10 @@ GEM
       unicode-display_width (~> 1.0, >= 1.0.1)
     ruby-progressbar (1.8.1)
     sass (3.4.22)
+    sassc (1.10.1)
+      bundler
+      ffi (~> 1.9.6)
+      sass (>= 3.3.0)
     unicode-display_width (1.1.1)
 
 PLATFORMS
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -8,7 +8,7 @@ PATH
       jshintrb (~> 0.3.0)
       json
       sass
-      sassc
+      sassc (~> 1.11.0)
 
 GEM
   remote: https://rubygems.org/
@@ -55,8 +55,8 @@ GEM
       ruby-progressbar (~> 1.7)
       unicode-display_width (~> 1.0, >= 1.0.1)
     ruby-progressbar (1.8.1)
-    sass (3.4.22)
-    sassc (1.10.1)
+    sass (3.4.23)
+    sassc (1.11.0)
       bundler
       ffi (~> 1.9.6)
       sass (>= 3.3.0)
--- a/lib/zendesk_apps_support/sass_functions.rb
+++ b/lib/zendesk_apps_support/sass_functions.rb
@@ -1,4 +1,25 @@
 # frozen_string_literal: true
+require 'sass'
+
+module Sass::Script::Functions
+  module AppAssetUrl
+    def app_asset_url(name)
+      assert_type name, :String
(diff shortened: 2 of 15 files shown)


lib/zendesk_apps_support/sass_functions.rb
```diff
@@ -1,12 +1,12 @@
 # frozen_string_literal: true
-require 'sass'
+require 'sassc'
 
-module Sass::Script::Functions
+module SassC::Script::Functions
   module AppAssetUrl
     def app_asset_url(name)
-      assert_type name, :String
+      raise ArgumentError, "Expected #{name} to be a string" unless name.is_a? Sass::Script::Value::String
```
sandlerr: type checking from https://github.com/sass/sassc-ruby/blob/master/test/functions_test.rb#L223

lib/zendesk_apps_support/stylesheet_compiler.rb
```diff
@@ -11,7 +11,7 @@ def initialize(source, app_id, url_prefix)
     end
 
     def compile
-      Sass::Engine.new(wrapped_source, syntax: :scss, app_asset_url_builder: self).render
+      SassC::Engine.new(wrapped_source.dup, syntax: :scss, app_asset_url_builder: self).render
```
sandlerr: was getting errors mutating frozen strings here, not entirely sure why.

sandlerr:
Not sure if we need the second commit or not, it'll save a few bytes and most debugging is done with devtools rather than inspecting style tags.

spec/validations/stylesheets_spec.rb
```diff
@@ -29,6 +29,6 @@
 
     errors = ZendeskAppsSupport::Validations::Stylesheets.call(package)
     expect(errors.count).to eq(1)
-    expect(errors.first.to_s).to match(/Sass error: Invalid CSS after.*/)
+    expect(errors.first.to_s).to match(/Sass error: Error: Invalid CSS after.*/)
```
adammw: hmm, should we remove the second "error:" part?

adammw:
Are there likely to be enough differences that we'll see some apps failing to compile? Should we put it behind Arturo rollout?

sandlerr:
Got it working on windows, although the setup is a little more complicated than just `gem install zendesk_apps_tools` like before.

1. install the dev kit https://rubyinstaller.org/downloads/
2. `gem install ffi --platform=ruby` - this should have happened automatically but I needed to trigger it
3. `gem install sassc` - optional debug step to test that the compilation succeeds here as well
4. `gem install zendesk_apps_tools`

sandlerr:
> Are there likely to be enough differences that we'll see some apps failing to compile? Should we put it behind Arturo rollout?

It's not impossible. This pulls in v3.3.6 of libsass => https://github.com/sass/libsass/blob/master/docs/compatibility-plan.md#33

v3.4.0 was released with some fixes => https://github.com/sass/libsass/releases

@svizzari @danielbreves what do you think?

Gemfile.lock
```diff
@@ -54,6 +55,10 @@ GEM
       unicode-display_width (~> 1.0, >= 1.0.1)
     ruby-progressbar (1.8.1)
     sass (3.4.22)
```
svizzari: sass is installed at 3.4 here, but 3.3 under sassc?  What is the current supported version before this PR, 3.?

sandlerr: libsass 3.3 still targets compatibility with sass 3.4. It appears to be reasonably close. I don't think we document any particular version of the sass spec: https://developer.zendesk.com/apps/docs/agent/styling

adammw:
Hmm, anything we can do to make the Windows setup easier? I can't see this going down easily if it makes it harder to install ZAT

Gemfile.lock
svizzari: The docs we link to are linked to a gem version, so I _guess_ it's implicit (if not necessarily true from our perspective)?

There wouldn't seem to be any harm in arturo'ing, we only need to consume in ZAM(S), so it's more straightforward

sandlerr: > The docs we link to are linked to a gem version, so I guess it's implicit (if not necessarily true from our perspective)?

Where's that? I couldn't find any reference to a specific sass version.

I'll make it arturoable.

sandlerr: actually, shouldn't this be a manifest experimental option? @svizzari

sandlerr:
@zendesk/vegemite ready for review

Gemfile.lock
adammw: I can see arguments either way - if it's a security-related change, not sure it should be up to the app to opt-in, but in the other case, it makes more sense if we provide the ability to do both before we force app developers to do so eventually.

spec/package_spec.rb
```diff
@@ -65,6 +65,13 @@
       expect(js).to eq(expected)
     end
 
+    it 'should generate css with the new flag (I know right, look at that function name)' do
```
adammw: :cough: :cough:... could we not?

sandlerr:
@adammw @zendesk/vegemite 💇‍♂️

sandlerr:
@svizzari since https://github.com/sass/sassc-ruby/pull/57 was merged this is now libsass 3.4 so should have pretty good compatibility

spec/package_spec.rb
adammw: i was really only asking not to call attention to it in the spec name, but 👍

PR #150, opened 2016-12-15
[APPS-3271] Ensure requirements only apps specify Support as their product

adammw:
:v:

/cc @zendesk/vegemite

### Description
Ensure requirements only apps specify Support as their product. This is to pass validations that the app can be uploaded from within Support.

### References
* JIRA: https://zendesk.atlassian.net/browse/APPS-3271

### Risks
* low: breaks product-related validations / features

### DIFF
--- a/lib/zendesk_apps_support/manifest.rb
+++ b/lib/zendesk_apps_support/manifest.rb
@@ -47,9 +47,13 @@ module ZendeskAppsSupport
 
     def products
       @products ||=
-        location_options.map { |lo| lo.location.product_code }
-                        .uniq
-                        .map { |code| Product.find_by(code: code) }
+        if requirements_only?
+          [ Product::SUPPORT ]
+        else
+          location_options.map { |lo| lo.location.product_code }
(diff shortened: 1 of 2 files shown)


sandlerr:
is it possible to add a regression test? does this affect all requirements apps?

dpawluk:
Will this break channel framework and requirements apps based on current docs?

sandlerr:
No, this is the fix for currently-broken requirements only apps.

danielbreves:
@sandlerr we had a browser test for this, but it was disabled, @token-cjg submitted a PR to fix it: https://github.com/zendesk/zendesk_browser_tests/pull/3720

sandlerr:
this seems unit testable, why would we not have one? With a tight merge / qa / deploy cycle those tests might not pick it up before it hits production anyway.

adammw:
By unit testable, do you mean for the old behaviour? Originally I left it out since we were planning for a production patch, however we can probably add it if we have time now.

PR #152, opened 2016-12-21
[AF-572] Implement validation to prevent the creation of v1 apps targeting v2 only locations

fvzzy:
:v:

/cc @zendesk/vegemite

### Description
Adds validation for v2 only locations in v1 apps

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-572

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -97,8 +97,6 @@ en:
               version.
             locations_cant_be_urls: App locations must not be URLs for this framework
               version.
-            locations_not_in_product: App locations must be valid for the selected
-              product.
             invalid_locations_for_version: App locations are not valid for this framework
               version.
         warning:
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -97,6 +97,10 @@ en:
               version.
             locations_cant_be_urls: App locations must not be URLs for this framework
               version.
+            locations_not_in_product: App locations must be valid for the selected
+              product.
+            invalid_locations_for_version: App locations are not valid for this framework
+              version.
         warning:
           app_build:
             deprecated_version: You are targeting a deprecated version of the framework.
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -97,8 +97,8 @@ en:
               version.
             locations_cant_be_urls: App locations must not be URLs for this framework
               version.
-            invalid_locations_for_version: App locations are not valid for this framework
-              version.
+            invalid_v1_location: "%{invalid_locations} is an invalid location in framework
+              v1."
         warning:
           app_build:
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -97,6 +97,8 @@ en:
               version.
             locations_cant_be_urls: App locations must not be URLs for this framework
               version.
+            invalid_v1_location: "%{invalid_locations} is an invalid location in framework
```
sandlerr: need to support one or many? https://github.com/zendesk/zendesk_apps_support/blob/e16fdfb6fe5fa4111845db82d27c5fd28d6ade1d/config/locales/en.yml#L10-L11

lib/zendesk_apps_support/manifest.rb
```diff
@@ -80,8 +80,11 @@ def app_class_properties
     end
 
     def unknown_locations(host)
+      product = Product.find_by(name: host)
+
       if locations.key?(host)
-        locations[host].keys.uniq - Location::LOCATIONS_AVAILABLE.map(&:name)
+        product_locations = Location.where(product_code: product.code)
+        locations[host].keys.uniq - product_locations.map(&:name)
```
sandlerr: :+1: nice

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -178,6 +179,25 @@ def invalid_location_error(package)
           errors
         end
 
+        def invalid_v1_location(package)
+          if package.manifest.framework_version &&
```
sandlerr: mind removing a layer of nesting by returning early?

```ruby
def invalid_v1_location(package)
  return unless package.manifest.framework_version && Gem::Version ...

  invalid_locations = []
  ...
end
```

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -178,6 +179,25 @@ def invalid_location_error(package)
...
+            Gem::Version.new(package.manifest.framework_version) < Gem::Version.new('2')
+
+            invalid_locations = []
```
sandlerr: code golf: could write it as

```ruby
invalid_locations = package.manifest.location_options
  .map(&:location)
  .compact
  .select(&:v2_only)
  .map(&:name)
```

sandlerr:
does setting iframe URLs in a v1 app also error out?

adammw:
I might be overengineering here, but having a 'v2' and 'v1' variables seems to smell of repetition a bit to me, would it be worth having some whitelist approach in case we add more locations for future versions?

injuhwang:
@zendesk/i18n

PR #153, opened 2016-12-23
[AF-425] Set products for no location apps

adammw:
:v:

/cc @zendesk/vegemite

### Description

Hard-codes support for Zendesk Support product for `marketingOnly` apps (similar to #150), and adds validation that a marketing only app must not be private.

For reference, the old PR is here if anyone wants it: https://github.com/zendesk/zendesk_apps_support/compare/master...adammw/products-for-no-location-apps-bak

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-425

### Risks
* low: could break products for marketing only apps (which don't use ZAS until https://github.com/zendesk/zendesk_app_market/pull/2000 anyway)

### DIFF
--- a/lib/zendesk_apps_support/manifest.rb
+++ b/lib/zendesk_apps_support/manifest.rb
@@ -47,7 +47,7 @@ module ZendeskAppsSupport
 
     def products
       @products ||=
-        if requirements_only?
+        if requirements_only? || marketing_only?
           [ Product::SUPPORT ]
         else
           location_options.map { |lo| lo.location.product_code }
--- a/lib/zendesk_apps_support/manifest.rb
+++ b/lib/zendesk_apps_support/manifest.rb
@@ -50,9 +50,7 @@ module ZendeskAppsSupport
         if requirements_only? || marketing_only?
           [ Product::SUPPORT ]
         else
-          location_options.map { |lo| lo.location.product_code }
-                          .uniq
-                          .map { |code| Product.find_by(code: code) }
+          products_from_locations
         end
     end
 
@@ -133,6 +131,12 @@ module ZendeskAppsSupport
     private
 
     attr_reader :locations
+
+    def products_from_locations
+      location_options.map { |lo| lo.location.product_code }
+                      .uniq
+                      .map { |code| Product.find_by(code: code) }
+    end
 
     def set_locations_and_hosts
       @locations =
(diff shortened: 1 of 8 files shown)


lib/zendesk_apps_support/manifest.rb
```diff
@@ -115,7 +117,9 @@ def initialize(manifest_text)
         instance_variable_set(:"@#{ruby}", m[json])
       end
       @requirements_only ||= false
+      @requirements_only = false if @requirements_only.is_a?(Array) && @requirements_only.empty?
```
sandlerr: why would these be arrays?

adammw: updated description. although maybe i should also support strings too.
I did mention I'm not 100% sure this is the best way to go (i'd almost prefer a "requirements" location), but it's definitely the easiest to implement

spec/manifest_spec.rb
```diff
@@ -46,7 +47,7 @@
       experiments: {
         hashParams: true
       }
-    }
+    }.dup
```
danielbreves: ❓

adammw: I thought I was seeing the hash get cached between tests. I'm not sure if that was the case or not, but seems pretty harmless

adammw: actually i should remove it, pretty sure it's useless

lib/zendesk_apps_support/manifest.rb
sandlerr: :+1: on supporting strings too

adammw:
Added string support and validations for passing an array into requirementsOnly/marketingOnly properties - only other thing I could think of is potentially whitelisting which products can accept marketing or requirements only apps?

adammw:
^^ /cc @zendesk/wombat

lib/zendesk_apps_support/manifest.rb
```diff
@@ -154,6 +168,12 @@ def set_locations_and_hosts
         end
     end
 
+    def normalize_array_or_string_to_array_or_bool(val)
+      val = [ val ] if val.is_a?(String)
+      val = false if val.is_a?(Array) && val.empty?
```
svizzari: Could the first line return explicitly if val was a string rather than falling through to be reevaluated here?

adammw:
:haircut: @svizzari

adammw:
:haircut: Updated code & description to avoid controversial product specification and just hardcode it to the only product we support marketing only apps on anyway.

PR #154, opened 2017-01-04
make some stuff public and cache app_js

sandlerr:
cc @zendesk/vegemite

### DIFF
--- a/lib/zendesk_apps_support/package.rb
+++ b/lib/zendesk_apps_support/package.rb
@@ -191,21 +191,18 @@ module ZendeskAppsSupport
       ''
     end
 
+    def app_js
+      if @is_cached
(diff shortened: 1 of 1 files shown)


lib/zendesk_apps_support/package.rb
```diff
@@ -191,22 +191,19 @@ def app_css
...
-    def runtime_translations(translations)
-      result = translations.dup
-      result.delete('name')
-      result.delete('description')
-      result.delete('long_description')
-      result.delete('installation_instructions')
-      result
-    end
-
     def templates
```
adammw: should these be cached as well?

sandlerr: will do in a followup

PR #155, opened 2017-01-05
some cleanup and type check the manifest

sandlerr:
### Description

Does a lot of what hash_schema was used for in public app validation - we should be doing this for all apps

cc @zendesk/vegemite 

### Risks

- [medium] failing validations for manifest files that should be considered valid


config/locales/translations/zendesk_apps_support.yml
```diff
@@ -57,6 +57,18 @@ parts:
       title: "App builder job: an experiment has been specifed that is not publically available or otherwise invalid. https://en.wikipedia.org/wiki/Manifest_file"
       value: "Manifest specifies unknown or unavailable experiment: %{experiment}"
   - translation:
+      key: "txt.apps.admin.error.app_build.unacceptable_hash"
+      title: "App builder job: the value for a field needs to be key-value pairs, but it was something else. https://en.wikipedia.org/wiki/Manifest_file"
+      value: "The manifest field %{field} needs to be key-value pairs with string keys, but it was %{value}"
```
sandlerr: good point, we don't have consistency around that for the other strings.

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -59,7 +57,56 @@ def collate_manifest_errors(package)
...
+              field: 'parameters',
+              value: parameters.class.to_s
+            )
+          end
+          errors
+        end
+
+        def string_error(manifest)
+          errors = []
+          manifest_strings = %i(
```
adammw: could these live in or near the RUBY_TO_JSON constant in the other manifest.rb?

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -59,7 +57,56 @@ def collate_manifest_errors(package)
...
+            default_locale
+            version
+            framework_version
+            remote_installation_url
+            terms_conditions_url
+            google_analytics_code
+          )
+          errors.concat(manifest_strings.map do |field|
```
adammw: finding this a bit ugly/unreadable. could do:
```
errors << manifest_strings.map do |field|
  validate_string(manifest.public_send(field), field)
end
```

and don't worry about the .compact, (since collate_manifest_errors does that anyway) or do it at the end of the function

spec/validations/marketplace_spec.rb
```diff
@@ -38,7 +38,7 @@
       end
     end
 
-    context 'with whitelisted experiements' do
+    context 'with whitelisted experiments' do
```
adammw: ty

adammw:
reading this I can value what a "schema" gives us (since you don't need to duplicate which field should be what), I think the issue was more in the implementation and complexity of what we had. this is a good start at least and should do for now though 👍

sandlerr:
yep, would be good to combine the type information with the ruby_to_json map, we can think of a neat way to do it later, preferably without throwing too many deserialization errors since that limits us to only one error per attempt.

lib/zendesk_apps_support/validations/manifest.rb
sandlerr: won't that need to be flattened?

lib/zendesk_apps_support/validations/manifest.rb
sandlerr: in two minds about this, I like to put the usage near the definition

adammw: ok, put it off until we think about how we might integrate them later

lib/zendesk_apps_support/validations/manifest.rb
adammw: it is, isn't it?

sandlerr: collate_manifest_errors flattens and compacts, that should do I think

PR #157, opened 2017-01-23
[AF-402] Add rollbar_zaf_access_token

danielbreves:
✌️ 

### References
- JIRA: https://zendesk.atlassian.net/browse/AF-402

/cc @zendesk/vegemite


lib/zendesk_apps_support/assets/installed.js.erb
```diff
@@ -14,4 +14,6 @@
   <% end %>
 }());
 
+ZendeskApps.rollbarAccessToken = "<%= rollbar_zaf_access_token %>";
```
sandlerr: nitpick: can this be an empty line in the template if the token is empty?

lib/zendesk_apps_support/assets/installed.js.erb
```diff
@@ -14,4 +14,8 @@
   <% end %>
 }());
 
+<% unless rollbar_zaf_access_token.empty? %>
```
sandlerr: sorry, need to guard against this too:

```ruby
% irb
irb(main):001:0> nil.empty?
NoMethodError: undefined method `empty?' for nil:NilClass
	from (irb):1
	from /Users/rsandler/.rbenv/versions/2.4.0/bin/irb:11:in `<main>'
```

danielbreves: Why would it be `nil`?

sandlerr: if this code goes live server-side before the variable is available, or if running in zat

danielbreves: @sandlerr it defaults to an `empty` string [here](https://github.com/zendesk/zendesk_apps_support/pull/157/files#diff-67a579144c0a408fe3168e4a92bf4420R19).

sandlerr: oops didn't notice that

PR #158, opened 2017-03-01
Let's log a critical error with rollbar

ocke:
:v:

/cc @zendesk/vegemite

### Description
In the aftermath of the S3 cache. Let's log a critical error if the app isn't loaded.

### References
* JIRA: 

### Risks

### DIFF
--- a/lib/zendesk_apps_support/assets/installed.js.erb
+++ b/lib/zendesk_apps_support/assets/installed.js.erb
@@ -6,6 +6,8 @@
   <% installations.each do |installation| %>
     if (ZendeskApps[<%= installation.app_name.to_json %>]) {
       ZendeskApps[<%= installation.app_name.to_json %>].install(<%= installation.to_json %>);
+    } else {
+      Rollbar && Rollbar.critical("App code missing: '<%= installation.app_name.to_json %>' (<%= installation.app_id.to_json %>)");
     }
   <% end %>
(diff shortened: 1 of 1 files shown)


lib/zendesk_apps_support/assets/installed.js.erb
```diff
@@ -6,6 +6,8 @@
   <% installations.each do |installation| %>
     if (ZendeskApps[<%= installation.app_name.to_json %>]) {
       ZendeskApps[<%= installation.app_name.to_json %>].install(<%= installation.to_json %>);
+    } else {
+      Rollbar.critical("App code missing: '<%= installation.app_name.to_json %>' (<%= installation.app_id.to_json %>)");
```
sandlerr: won't this just error out in Chat? not sure we have rollbar there.

fvzzy: I thought we (also) wanted this error going to Datadog?

ocke: oh ok, so `Rollbar && `?

ocke: btw, i found the errors. they show up in `lotus`

sandlerr:
we need to make the errors should show up in zaf, not in lotus.

PR #159, opened 2017-03-06
Add organization fields as a resource type in app requirements

fvzzy:
:v:

/cc @zendesk/vegemite

### Description
Add organization fields as a resource type in app requirements

### References
* JIRA: Related to testing https://zendesk.atlassian.net/browse/AF-577

### Risks
* [low] breaks validations for app requirements

### DIFF
--- a/lib/zendesk_apps_support/validations/requirements.rb
+++ b/lib/zendesk_apps_support/validations/requirements.rb
@@ -56,7 +56,7 @@ module ZendeskAppsSupport
           user_fields = requirements['user_fields'] || {}
           organization_fields = requirements['organization_fields'] || {}
           custom_fields = user_fields.merge(organization_fields)
-          return unless custom_fields
+          return if custom_fields.empty?
           [].tap do |errors|
             custom_fields.each do |identifier, fields|
               next if fields.include? 'key'
--- a/lib/zendesk_apps_support/validations/requirements.rb
+++ b/lib/zendesk_apps_support/validations/requirements.rb
@@ -53,14 +53,15 @@ module ZendeskAppsSupport
         end
 
         def invalid_custom_fields(requirements)
-          user_fields = requirements['user_fields'] || {}
-          organization_fields = requirements['organization_fields'] || {}
-          custom_fields = user_fields.merge(organization_fields)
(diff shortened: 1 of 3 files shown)


lib/zendesk_apps_support/validations/requirements.rb
```diff
@@ -52,11 +52,13 @@ def excessive_requirements(requirements)
...
         end
 
-        def invalid_user_fields(requirements)
-          user_fields = requirements['user_fields']
-          return unless user_fields
+        def invalid_custom_fields(requirements)
+          user_fields = requirements['user_fields'] || {}
+          organization_fields = requirements['organization_fields'] || {}
+          custom_fields = user_fields.merge(organization_fields)
+          return unless custom_fields
```
sandlerr: this will never return

lib/zendesk_apps_support/validations/requirements.rb
```diff
@@ -52,11 +52,13 @@ def excessive_requirements(requirements)
           ValidationError.new(:excessive_requirements, max: MAX_REQUIREMENTS, count: count) if count > MAX_REQUIREMENTS
         end
 
-        def invalid_user_fields(requirements)
-          user_fields = requirements['user_fields']
-          return unless user_fields
+        def invalid_custom_fields(requirements)
+          user_fields = requirements['user_fields'] || {}
+          organization_fields = requirements['organization_fields'] || {}
+          custom_fields = user_fields.merge(organization_fields)
```
sandlerr: not sure about the logic here, is it possible for an org field to clobber a user field? can't we just iterate over the two hashes separately?

spec/validations/requirements_spec.rb
```diff
@@ -79,6 +79,43 @@
     end
   end
 
+  context 'a user field requirement is lacking required fields' do
```
sandlerr: how about `a user field requirement lacking a "key" field`

sandlerr: or leave it, same as other tests I guess 🤷‍♂️

sandlerr:
@fvzzy this needs to be tagged and released so we can start rolling it out

PR #160, opened 2017-03-27
Fix cheeky ruby code. Duplicate objects

fbvilela:
@zendesk/vegemite @zendesk/wombat 

### Description

Fixes ruby code responsible for duplicate keys


fbvilela:
I've done a version bump and a release already.

PR #164, opened 2017-06-15
[OFFAPPS-814] case sensitive locale check

ocke:
:v:

/cc @zendesk/vegemite

### Description
currently zat doesn't check case sensitive. And it doesn't check if we actually have that locale.

There is one downside of doing this. If we add a local we have to update ZAS (and ZAT/ZAM) before we can use it.

### References
* JIRA: https://zendesk.atlassian.net/browse/OFFAPPS-814

### Risks
[high] a lot of apps will fail this check (correctly!).

### DIFF
--- a/lib/zendesk_apps_support/validations/translations.rb
+++ b/lib/zendesk_apps_support/validations/translations.rb
@@ -8,7 +8,7 @@ module ZendeskAppsSupport
       TRANSLATIONS_PATH = %r{^translations/(.*)\.json$}
                           # from https://support.zendesk.com/api/v2/locales/agent.json
                           # manually added 'en'
-      VALID_LOCALES     = %w{bg cs da de en en-CA en-GB en-US es es-419 es-ES fi fr fr-CA it ja ko nl no pl pt pt-BR ro ru sv tr uk zh-CN zh-TW}.freeze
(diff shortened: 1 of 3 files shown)


lib/zendesk_apps_support/validations/translations.rb
```diff
@@ -6,7 +6,9 @@ module ZendeskAppsSupport
   module Validations
     module Translations
       TRANSLATIONS_PATH = %r{^translations/(.*)\.json$}
-      VALID_LOCALE      = /^[a-z]{2}(-\w{2,3})?$/
+                          # from https://support.zendesk.com/api/v2/locales/agent.json
+                          # manually added 'en'
+      VALID_LOCALES     = %w{bg cs da de en en-CA en-GB en-US es es-419 es-ES fi fr fr-CA it ja ko nl no pl pt pt-BR ro ru sv tr uk zh-CN zh-TW}.freeze
```
sandlerr: can we just hit the endpoint?

ocke: yes i thought aobut that, but then I thought, do we really want to build in a 2 sec delay everytime we validate?

PR #166, opened 2017-06-20
[AF-681] Add SUNSETTING and SUNSETTED

nick-f:
:v:

/cc @zendesk/vegemite

### Description
Add support for arrays of versions as well as `sunsetting` and `sunsetted` versions.

This can be used to display warnings in ZAT, such as when running [`zat update`](https://github.com/zendesk/zendesk_apps_tools/pull/190).

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-681

### Risks
* [low] New methods being added

### DIFF
--- a/lib/zendesk_apps_support/app_version.rb
+++ b/lib/zendesk_apps_support/app_version.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 module ZendeskAppsSupport
-  # At any point in time, we support up to four types of versions:
+  # At any point in time, we support up to four versions:
   #  * deprecated -- we will still serve apps targeting the deprecated version,
   #                  but newly created or updated apps CANNOT target it
   #  * sunsetting -- we will soon be removing support for this version;
@@ -10,15 +10,14 @@ module ZendeskAppsSupport
   #  * future     -- we will serve apps targeting the future version;
   #                  newly created or updates apps MAY target it, but it
   #                  may change without notice
-  # Versions can be listed as strings or in arrays
   class AppVersion
     DEPRECATED = '0.5'
     SUNSETTING = '1.0'
     CURRENT    = '2.0'
     FUTURE     = nil
 
-    TO_BE_SERVED     = [DEPRECATED, SUNSETTING, CURRENT, FUTURE].compact.flatten.freeze
-    VALID_FOR_UPDATE = [SUNSETTING, CURRENT, FUTURE].compact.flatten.freeze
+    TO_BE_SERVED     = [DEPRECATED, SUNSETTING, CURRENT, FUTURE].compact.freeze
+    VALID_FOR_UPDATE = [SUNSETTING, CURRENT, FUTURE].compact.freeze
 
     attr_reader :current
 
@@ -38,11 +37,11 @@ module ZendeskAppsSupport
     end
 
     def deprecated?
-      DEPRECATED.include?(@version)
+      @version == DEPRECATED
     end
 
     def sunsetting?
-      SUNSETTING && SUNSETTING.include?(@version)
+      @version == SUNSETTING
     end
 
     def obsolete?
--- a/lib/zendesk_apps_support/app_version.rb
+++ b/lib/zendesk_apps_support/app_version.rb
@@ -22,8 +22,7 @@ module ZendeskAppsSupport
     attr_reader :current
 
     def initialize(version)
-      @version = version.to_s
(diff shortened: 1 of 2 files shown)


lib/zendesk_apps_support/app_version.rb
```diff
@@ -31,7 +35,15 @@ def valid_for_update?
     end
 
     def deprecated?
-      @version == DEPRECATED
+      DEPRECATED.include?(@version)
+    end
+
+    def sunsetting?
+      SUNSETTING && SUNSETTING.include?(@version)
```
ocke: you forgot the leading `true &&`

ocke: Isn't `SUNSETTING` always true?

lib/zendesk_apps_support/app_version.rb
```diff
@@ -31,7 +35,15 @@ def valid_for_update?
...
+    end
+
+    def current
```
ocke: `attr_reader` ? @danielbreves

lib/zendesk_apps_support/app_version.rb
nick-f: At the moment, yes, but this is preparing for when there are no versions currently being sunset which will happen sometime in the future.

lib/zendesk_apps_support/app_version.rb
nick-f: 💇‍♂️

lib/zendesk_apps_support/app_version.rb
```diff
@@ -1,24 +1,31 @@
...
-    TO_BE_SERVED     = [DEPRECATED, CURRENT, FUTURE].compact.freeze
-    VALID_FOR_UPDATE = [CURRENT, FUTURE].compact.freeze
+    TO_BE_SERVED     = [DEPRECATED, SUNSETTING, CURRENT, FUTURE].compact.flatten.freeze
+    VALID_FOR_UPDATE = [SUNSETTING, CURRENT, FUTURE].compact.flatten.freeze
+
+    attr_reader :current
 
     def initialize(version)
       @version = version.to_s
       @version.freeze
```
fvzzy: nitpick: could be chained (?)

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -292,7 +292,7 @@ def missing_framework_version(manifest)
         end
 
         def invalid_version_error(manifest, package)
-          valid_to_serve = AppVersion::TO_BE_SERVED
+          valid_to_serve = AppVersion::TO_BE_SERVED.compact
```
fvzzy: Isn't this const already `compact`ed in the class?

lib/zendesk_apps_support/app_version.rb
```diff
@@ -31,7 +38,11 @@ def valid_for_update?
     end
 
     def deprecated?
-      @version == DEPRECATED
+      DEPRECATED.include?(@version)
```
danielbreves: Why did you need this change?

lib/zendesk_apps_support/validations/manifest.rb
danielbreves: Isn't `TO_BE_SERVED` already `compact`?

nick-f: Good pickup. Fixed.

lib/zendesk_apps_support/app_version.rb
nick-f: Allows for `DEPRECATED` to be an array of values. Or, when 1.0 is deprecated, will we just drop 0.5 from the list entirely? If the latter is true, then this can be removed.

lib/zendesk_apps_support/validations/manifest.rb
nick-f: Indeed it is, good pickup. Fixed.

lib/zendesk_apps_support/app_version.rb
danielbreves: Considering this is a public constant I think it would be dangerous to change its type. See https://github.com/zendesk/zendesk_app_market/search?utf8=%E2%9C%93&q=AppVersion&type=.

nick-f: Ah, I see. Reverted back to how it was before. 💇‍♂️

PR #167, opened 2017-07-04
[APPS-3371] Add a new validation method and distinguish create from update

fbvilela:
@zendesk/wombat @zendesk/vegemite 

### Description
Adding a new validation method to be used in the zendesk_app_market.
Sunsetting will mean updates only and no new apps.

once this gem is bumped, I will update the app market and then write specs for it.

### References
* JIRA: https://zendesk.atlassian.net/browse/APPS-3371

### Risks
low - breaks app updates/create

### DIFF
--- a/lib/zendesk_apps_support/app_version.rb
+++ b/lib/zendesk_apps_support/app_version.rb
@@ -4,7 +4,7 @@ module ZendeskAppsSupport
   #  * deprecated -- we will still serve apps targeting the deprecated version,
   #                  but newly created or updated apps CANNOT target it
   #  * sunsetting -- we will soon be removing support for this version;
-  #                  newly created or updated apps SHOULD target the current version
+  #                  updated apps SHOULD target the current version, no new apps allowed
   #  * current    -- we will serve apps targeting the current version;
   #                  newly created or updated apps SHOULD target it
   #  * future     -- we will serve apps targeting the future version;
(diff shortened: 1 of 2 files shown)


nick-f:
At the moment, ZAT uses the sunsetting version to check whether to display the sunsetting message.
Based on the new description of what sunsetting should be, shouldn't the sunsetting value be nil until August 1st? (If so, we'd need to update ZAT to deal with that.)

fbvilela:
Nick, I was thinking to make those changes and not merge it until August 1st.
Should we discuss what's the best strategy? Maybe Phil would be happy to end new v1 apps right now?

nick-f:
@fbvilela I highly doubt we'd be able to end new v1 apps now as the August 1st date is what's been communicated for a while to developers.

Not merging it until August works for me (ZAT is working now and will work with those changes too) so as long as ZAM is happy to wait for it.

fbvilela:
@zendesk/vegemite @zendesk/wombat . This is ready for review. 
I have a branch on the zendesk_app_market project ready to bump this gem and fix failing tests.

The idea is to merge THIS PR before august 1st, bump zam and get it deployed on the 1st.

xuebingli:
A spec would be great.

fbvilela:
@xuebingli 💇‍♂️

PR #168, opened 2017-07-25
[AF-776] Check for existence of app name in en.json

nick-f:
:v:

/cc @zendesk/vegemite

### Description
Following on from recent changes that require short_description, long_description, and installation_instructions for marketplace apps, en.json is also required to have the app > name property.

```
app {
    name: "This is now required"
}
```

ZAT validation uses ZAS. When running `zat validate` a message is now displayed when a required property is missing.

A [PR to bump ZAT's version of ZAS will be required](https://github.com/zendesk/zendesk_apps_tools/pull/198) once this is merged and tagged.

image

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-776

### Risks
* [low] Apps validations will fail

### DIFF
--- a/lib/zendesk_apps_support/validations/translations.rb
+++ b/lib/zendesk_apps_support/validations/translations.rb
@@ -17,6 +17,7 @@ module ZendeskAppsSupport
             path_match = TRANSLATIONS_PATH.match(file.relative_path)
             if path_match
               errors << locale_error(file, path_match[1]) << json_error(file)
+              errors << required_keys(file) if errors.compact.empty?
             end
           end.compact
         end
@@ -45,6 +46,14 @@ module ZendeskAppsSupport
           end
         rescue JSON::ParserError => e
           ValidationError.new('translation.not_json', file: file.relative_path, errors: e)
+        end
+
+        def required_keys(file)
+          return if file.relative_path != 'translations/en.json' || JSON.parse(file.read)['app']['name']
+
+          ValidationError.new('translation.missing_required_key',
+                              file: file.relative_path,
+                              missing_key: 'app.name')
         end
 
         def validate_translation_format(json)
--- a/spec/validations/translations_spec.rb
+++ b/spec/validations/translations_spec.rb
@@ -55,6 +55,32 @@ describe ZendeskAppsSupport::Validations::Translations do
     end
   end
 
+  context 'when there is a en.json' do
+    context 'required keys are missing' do
+      let(:translation_files) do
+        [double('AppFile',
+                relative_path: 'translations/en.json',
+                read: read_fixture_file('invalid_en.json'),
+                to_s: 'translations/en.json')]
+      end
+
+      it 'should report the error' do
+        expect(subject.length).to eq(1)
+        expect(subject[0].to_s).to match(/Missing required key from/)
+      end
+    end
+
+    context 'required keys are present' do
+      let(:translation_files) do
+        [double('AppFile', relative_path: 'translations/en.json', read: read_fixture_file('valid_en.json'))]
+      end
+
+      it 'should be valid' do
+        expect(subject.length).to eq(0)
+      end
+    end
+  end
+
   context 'validate translation format when "package" is defined inside "app"' do
     context 'all the leaf nodes have defined "title" and "value"' do
       let(:translation_files) do
@@ -71,7 +97,7 @@ describe ZendeskAppsSupport::Validations::Translations do
         [double('AppFile', relative_path: 'translations/en-US.json', read: read_fixture_file('invalid_en-US.json'))]
       end
 
-      it 'should be invalid' do
+      it 'should report the error' do
         expect(subject.length).to eq(1)
         expect(subject[0].to_s).to match(/is invalid for translation/)
       end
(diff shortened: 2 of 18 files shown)


lib/zendesk_apps_support/validations/translations.rb
```diff
@@ -47,6 +47,15 @@ def json_error(file)
           ValidationError.new('translation.not_json', file: file.relative_path, errors: e)
         end
 
+        def required_keys(file)
+          return nil unless file.to_s == 'translations/en.json'
+
+          json = JSON.parse(file.read)
+          return nil if json['app']['name']
+
+          ValidationError.new('translation.missing_required_key', file: file.relative_path, missing_key: 'app.name')
```
fvzzy: So previously there was no validation against translations / there were no _required_ translation keys?

fvzzy: Also, should there be an additional test that demos this failure?

nick-f: Correct, previously there were no required translation keys.

I'll add some tests for it.

fbvilela: this is weird...The validation of translation keys already happen here: https://github.com/zendesk/apps_approval/blob/master/lib/apps_approval/validators/translations_validator.rb#L40

fbvilela: So, answering to @fvzzy 's question. There are current validations for translation keys in place

fvzzy: I think you might just need to add `name` into `apps_approval` then

nick-f: @fbvilela I was referring to what was happening in ZAS. I have no idea what happens in other places 😬 🙃

nick-f:
@fvzzy 💇‍♂️ 

@fbvilela 👀 ?

spec/validations/translations_spec.rb
```diff
@@ -58,7 +58,7 @@
   context 'validate translation format when "package" is defined inside "app"' do
     context 'all the leaf nodes have defined "title" and "value"' do
       let(:translation_files) do
-        [double('AppFile', relative_path: 'translations/en-US.json', read: read_fixture_file('valid_en.json'))]
+        [double('AppFile', relative_path: 'translations/en-US.json', read: read_fixture_file('valid_en-US.json'))]
```
fvzzy: Was there a reason you chose to create additional fixture files for these tests? For the purposes of the spec acting as documentation, in this case I would probably have just defined those _invalid_/_valid_ json structures as hashes inline with the rest of the spec

nick-f: I put them in separate files because they're a decent size and thought it was best to pull them out for readability.

spec/validations/translations_spec.rb
```diff
@@ -55,6 +55,32 @@
...
+  context 'when there is a en.json' do
+    context 'required keys are missing' do
+      let(:translation_files) do
+        [double('AppFile',
+                relative_path: 'translations/en.json',
+                read: read_fixture_file('invalid_en.json'),
+                to_s: 'translations/en.json')]
+      end
+
+      it 'should be invalid' do
```
fvzzy: nitpick: `should report the error`?

nick-f:
cc @zendesk/wombat

lib/zendesk_apps_support/validations/translations.rb
```diff
@@ -17,6 +17,7 @@ def call(package)
             path_match = TRANSLATIONS_PATH.match(file.relative_path)
             if path_match
               errors << locale_error(file, path_match[1]) << json_error(file)
+              errors << required_keys(file) if errors.compact.empty?
```
mtsolakiszen: Why are we only adding the error if there are no previous ones?

lib/zendesk_apps_support/validations/translations.rb
```diff
@@ -47,6 +48,17 @@ def json_error(file)
           ValidationError.new('translation.not_json', file: file.relative_path, errors: e)
         end
 
+        def required_keys(file)
+          return nil unless file.relative_path == 'translations/en.json'
```
mtsolakiszen: You don't need the `nil` here, can just be `return unless file.relative_path == 'translations/en.json'`.

Could even combine this with the guard below into something like:
`return if file.relative_path != 'translations/en.json' || JSON.parse(file.read)['app']['name']`

lib/zendesk_apps_support/validations/translations.rb
nick-f: Before this there was more duplication in `required_keys` of the guards that are in `json_error`. This is one way to avoid needing to check twice.

lib/zendesk_apps_support/validations/translations.rb
nick-f: @mtsolakiszen That's much nicer. 💇‍♂️

fbvilela:
💥 🥇 🏆

token-cjg:
Unit tests pass: check
Tested locally using zatdev alias as discussed: check

:+1: approved.

spec/validations/translations_spec.rb
```diff
@@ -55,10 +55,36 @@
...
+  context 'when there is a en.json' do
+    context 'required keys are missing' do
+      let(:translation_files) do
+        [double('AppFile',
+                relative_path: 'translations/en.json',
+                read: read_fixture_file('invalid_en.json'),
+                to_s: 'translations/en.json')]
+      end
+
+      it 'should report the error' do
```
token-cjg: 👏

PR #169, opened 2017-08-01
[CACT-272] Allow marketing only apps in chat

mtsolakiszen:
:bear:

### Description
Allow marketing only apps to have products and therefore be available for chat.

The products are read from the locations in the manifest and used even if there are no locations specified within the product hash.

If no products are specified it defaults to support to keep it backwards compatible.

### References
* JIRA: https://zendesk.atlassian.net/browse/CACT-272

### Risks
* [low] Breaks marketing only apps


fbvilela:
bump @zendesk/vegemite

PR #170, opened 2017-08-14
[APPS-3559] Add validation for multi-product marketplace content

seangoedecke:
:stop_sign: Do not merge this PR (& bump the gem) until the corresponding changes to the ZAM builder are made :stop_sign: 

Works on my machine.
Stuff this PR does:

1) Makes the keys `short_description`, `long_description` and `installation_instructions` mandatory in `en.json` (we'll accept `description` instead of `short_description`, for backwards compatibility) 

2) Allows these mandatory keys (& the other mandatory key, `name`), to be scoped by product, e.g.
```
{
  "app": {
  	"support": {
	    "name": "App Name (support)",
	    "short_description": "App description (support)",
	    "installation_instructions": "How to install the app (support)",
	    "long_description": "A longer version of the description (support)"
  	},
  	"chat": {  		
	    "name": "App Name (chat)",
	    "short_description": "App description (chat)",
	    "installation_instructions": "How to install the app (chat)",
	    "long_description": "A longer version of the description (chat)"
  	}
  }
}
```
(if there's no product scoping, we fall back to expecting all of these as top-level keys)

3) Ensures that the products given (as in the example above) match the products in the manifest

Jira: https://zendesk.atlassian.net/browse/APPS-3559


seangoedecke:
@zendesk/wombat @zendesk/vegemite Ready for review. Is making these keys mandatory going to break anything I should know about? It's part of the Jira so :man_shrugging:

fbvilela:
the mandatory keys already happens in `apps_approval`.  This is getting a bit confusing. Which gem is responsible for what. 

What's the reason for the mandatory keys in this PR?

seangoedecke:
@fbvilela Wherever else the validation has to go, it's got to go here, otherwise ZAT users will mistakenly think their apps are valid.

seangoedecke:
Looks like apps_approval uses ZAS anyway: https://github.com/zendesk/apps_approval/blob/e5ae401eaddf05f54d9f80a1b9c7764b7e8cf0b1/lib/apps_approval/app_validation.rb#L20

I reckon all new validations should just go here, so there's one source of truth. If apps_approval is more strict than ZAS, we should fix that.

xuebingli:
Since this PR introduces breaking changes, we should hold on merging this PR until the backend changes are merged and deployed.

lib/zendesk_apps_support/validations/translations.rb
```diff
@@ -7,6 +7,7 @@ module Validations
     module Translations
       TRANSLATIONS_PATH = %r{^translations/(.*)\.json$}
       VALID_LOCALE      = /^[a-z]{2}(-\w{2,3})?$/
+      MANDATORY_KEYS = %w(name description installation_instructions long_description).freeze
```
fbvilela: Sean, I think we are using short_description as mandatory in apps_approval ( and accepting description if present )

nick-f:
@seangoedecke @fbvilela There was some discussion about where to put things when I was adding mandatory keys for name in `en.json`. It's definitely confusing for us to know what validation happens where, between ZAT, ZAS, and apps_approval. 😖 

I was originally going to add checks for `short_description`, `long_description`, and `installation_instructions` as well but was told that ZAS should be checking for what private apps need (i.e. not care about those three, only care about name being present) and apps_approval checks for what public apps need.

Would it make sense to add some conditional checks here to only check for short, long, installation if the manifest says `private: false`? Seems a bit odd to fail private apps because of something that they will never use.

seangoedecke:
@nick-f Solid point. I've altered the code to keep `name` as the only mandatory key for private apps.

lib/zendesk_apps_support/validations/translations.rb
```diff
@@ -48,12 +48,56 @@ def json_error(file)
...
+          end
+          nil
+        end
 
-          ValidationError.new('translation.missing_required_key',
-                              file: file.relative_path,
-                              missing_key: 'app.name')
+        def validate_products_match_manifest_products(products, package, file_path)
+          manifest_products = package.manifest.products.map(&:name)
+          return if products.sort == manifest_products.sort
```
itirkarp: Can use subtraction instead of sort here too

itirkarp:
Is it worth booking some time together to make a decision about what should be in apps_validation and what in ZAS, even if it involves changing some stuff around? We can write it down somewhere for future reference

lib/zendesk_apps_support/validations/translations.rb
```diff
@@ -48,12 +48,56 @@ def json_error(file)
           ValidationError.new('translation.not_json', file: file.relative_path, errors: e)
         end
 
-        def required_keys(file)
-          return if file.relative_path != 'translations/en.json' || JSON.parse(file.read)['app']['name']
+        def validate_marketplace_content(file, package)
+          return [] if file.relative_path != 'translations/en.json'
```
seangoedecke: Sure, makes sense.

lib/zendesk_apps_support/validations/translations.rb
```diff
@@ -48,12 +48,56 @@ def json_error(file)
...
+            errors << validate_top_level_required_keys(json, file.relative_path)
+          else
+            errors << validate_products_match_manifest_products(present_product_keys, package, file.relative_path)
+            errors << validate_products_have_required_keys(json, present_product_keys, file.relative_path)
+          end
+          errors.compact
+        end
+
+        def validate_top_level_required_keys(json, file_path)
+          return if (json['app'].keys & MANDATORY_KEYS).sort == MANDATORY_KEYS.sort
```
seangoedecke: Oh yeah, I can just do `return if missing_keys.empty?`!

lib/zendesk_apps_support/validations/translations.rb
```diff
@@ -48,12 +48,56 @@ def json_error(file)
...
+          ValidationError.new(
+            'translation.missing_required_key',
+            file: file_path,
+            missing_key: missing_keys.join(', ')
+          )
+        end
+
+        def validate_products_have_required_keys(json, products, file_path)
+          products.each do |product|
+            next unless (json['app'][product].keys & MANDATORY_KEYS).sort != MANDATORY_KEYS.sort
```
seangoedecke: After the refactor I don't think this looks like as good an idea - it's just an array subtraction. I'd rather leave it as is.

lib/zendesk_apps_support/validations/translations.rb
seangoedecke: Nice, good catch. I've updated my code.

lib/zendesk_apps_support/validations/translations.rb
seangoedecke: Done. Now that we're accepting `description` as well as `short_description`, we need to do that anyway.

seangoedecke:
@zendesk/wombat Immediate upshot from our discussion: the validations in this PR can stay in ZAS, and this PR can be safely merged (once the changes to the builder are made). Time to provide thumbs?

PR #171, opened 2017-08-16
[CACT-314] Fix airbrake caused by unsafe hash check

mtsolakiszen:
:bear:

### Description
Airbrake was being caused by unsafe hash check when `app` key was not present.
This change makes the check for `app=>name` happen safely.

### Airbrake example
https://zendesk.airbrake.io/projects/105401/groups/2017693197944987628?tab=overview

### Risks
* [low] Doesn't validate presence of app name correctly

### DIFF
--- a/lib/zendesk_apps_support/validations/translations.rb
+++ b/lib/zendesk_apps_support/validations/translations.rb
@@ -49,7 +49,7 @@ module ZendeskAppsSupport
         end
 
         def required_keys(file)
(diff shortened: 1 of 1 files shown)


nick-f:
:+1: on ✅

itirkarp:
Yay no more airbrakes. Once this reaches prod please find all the relevant airbrakes and mark them resolved

alieu:
:+1:

PR #175, opened 2017-08-30
[CACT-284] Update the other place translations live in ZAS

seangoedecke:
I submitted a previous PR to validate that Chat only apps can't have requirements, but I only updated the error message in `en.yml`. Apparently the error Cactus gets comes from `zendesk_apps_support.yml`, which I didn't catch. This PR updates the other translation.

JIRA: https://zendesk.atlassian.net/browse/CACT-284


seangoedecke:
@zendesk/wombat @zendesk/vegemite Ready for review

PR #176, opened 2017-09-02
[LOCAL-44] Strings: reverse sweep (auto)

string-sweep:
/cc @zendesk/i18n @zendesk/i18n-dev

### Risks
Low. English edits and title improvements.


danielbreves:
@damiann tl;dr it's fixed now.

ZAS requires a rake command (`rake i18n:standardize`) to be run after updating translations to make them compatible with the i18n gem (we can't use the Zendesk one because this is an open-source project). We have a test that makes sure the generated file was updated after the zendesk one. Now because you only changed the zendesk one it failed. I had to make a manual change to the generated one to fix it.

PR #177, opened 2017-09-08
Fix requirements validations when invalid locations

mtsolakiszen:
:bear:

### Description
In the requirements validation a call has been added to `manifest.products`, however since this call happens during validation there's no guarantee that the locations are valid and this can error out.

I've replaced this call to use the `products_ignore_locations` method which only looks at the specified products without checking the locations and added some tests for when an array of support locations is used.

### Risks
* [low] Requirements validation issues

### DIFF
--- a/lib/zendesk_apps_support/manifest.rb
+++ b/lib/zendesk_apps_support/manifest.rb
@@ -56,6 +56,12 @@ module ZendeskAppsSupport
         end
     end
 
+    def products_ignore_locations
+      locations.keys.map do |product_name|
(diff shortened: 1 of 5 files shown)


token-cjg:
Tested zat against this branch

* removed from gemspec as dependency
* added relative path to gemfile `gem 'zendesk_apps_support', path: '../zendesk_apps_support'`
* ran `bundle install`
* ran `be rake spec`.  all tests passed
* ran `be rake cucumber`.  all tests passed

approved :+1:

PR #179, opened 2017-09-11
[AF-819] Sanitise SVGs during validation

fvzzy:
:v:

/cc @zendesk/vegemite

### Description
Sanitises malicious/malformed SVGs on the server using the gem [Loofah](https://github.com/flavorjones/loofah) during app submission. 
- Also raises a warning to the user when their app icon is regenerated (in the event the SVG contained harmless scripts/animations that the user expected to run)

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-819
* Ticket: https://support.zendesk.com/agent/tickets/2875295
* Investigation: https://docs.google.com/document/d/1gGU-AMmOV1KpZ5xFt8uW-W9AilwsA8VUdtMBN3dz3IA

### Risks
* [medium] overzealous sanitisation on poorly-formed (but otherwise harmless) SVGs causes icons to render incorrectly.


lib/zendesk_apps_support/validations/svg.rb
```diff
@@ -0,0 +1,28 @@
...
+    module Svg
+      class << self
+        def call(package)
+          errors = []
+          package.svg_files.each do |svg|
+            markup = svg.read
+            clean_markup = Loofah.scrub_xml_document(markup, :prune).to_html
+            filepath = svg.relative_path
+
+            next if clean_markup == markup
```
danielbreves: Is the `clean_markup` guaranteed to be unaltered in the case the svg is clean or could Loofah return something slightly reformatted?

spec/validations/svg_spec.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+require 'spec_helper'
+
+describe ZendeskAppsSupport::Validations::Svg do
+  let(:clean_markup) { "<svg viewBox=\"0 0 26 26\" id=\"zd-svg-icon-26-app\" width=\"100%\" height=\"100%\"><path fill=\"none\" stroke=\"currentColor\" stroke-linejoin=\"round\" stroke-width=\"2\" d=\"M4 8l9-5 9 5v9.7L13 23l-9-5.2zm9 5L4 8m9 5l9-5m-9 5v10\"></path></svg>\n" }
```
danielbreves: Any reason not to use single quotes here to avoid all the back-slashes? Could also use `%Q` to avoid the line break character. See https://ruby-doc.org/core-2.1.1/doc/syntax/literals_rdoc.html#label-Strings.

fvzzy: Rubocop doesn't like `%q` unless you're mixing double and single-quotes, but `%` cleans up the backslashes - thanks. Can't seem to work out how to get rid of the trailing line-break char though.

lib/zendesk_apps_support/validations/svg.rb
fvzzy: No, and this is where things could _potentially_ get hairy...

In the case of valid-but-less-than-perfect markup, Loofah will probably alter the contents slightly. I think this is _OK_ though, as the user does get a warning about this happening and should still see their SVG structure preserved and displayed as intended 🤷‍♂️ .

I might add a test case for this though, just for documentation purposes and to recognise that this (for now at least), is intended functionality.

alieu:
:+1: QA Approved

PR #180, opened 2017-09-14
[AF-819] Ensure leading XML declarations, leading doctypes and deprecated attributes don't trigger a rewrite on SVGs

fvzzy:
:v:

/cc @zendesk/vegemite

### Description
Ensures validation doesn't fail because of a leading XML declaration or DOCTYPE in an SVG, or because of the inclusion of deprecated (but otherwise innocuous) attributes e.g. the ones included the demo apps/fixtures for ZAT, ZAM etc.
- Picked up because of https://travis-ci.org/zendesk/zendesk_apps_tools/jobs/274888061.

### References
* Investigation: https://zendesk.atlassian.net/browse/AF-819
* Main story: https://zendesk.atlassian.net/browse/AF-838

### Risks
* [medium] test scenarios don't cover all potentials for markup, and could still trigger unnecessary rewrites on otherwise valid files.


lib/zendesk_apps_support/validations/svg.rb
```diff
@@ -22,6 +24,19 @@ def call(package)
...
+        private
+
+        # to ignore the optional XML declaration at the top of a document
+        def strip_declaration(markup)
+          Nokogiri::XML(markup).root.to_s
+        end
+
+        def are_equivalent_sans_declarations(clean_markup, markup)
+          # skip the check if it isn't possible for the markup to contain a declaration
+          return false unless Nokogiri::XML(markup).root.children.length >= 1
```
fvzzy: I realise the output of this Nokogiri operation isn't super obvious, but I spent a while [here](https://github.com/sparklemotion/nokogiri/wiki/Cheat-sheet) before I settled on this.

We could just regex for something that looked like `<xml?`... and strip that instead, but while I was working on it, it seemed less brittle to harness the capability of Nokogiri, since it's already imported in here.

lib/zendesk_apps_support/validations/svg.rb
```diff
@@ -22,6 +24,19 @@ def call(package)
...
         end
+
+        private
+
+        # to ignore the optional XML declaration at the top of a document
+        def strip_declaration(markup)
+          Nokogiri::XML(markup).root.to_s
+        end
+
+        def are_equivalent_sans_declarations(clean_markup, markup)
```
svizzari: Could the method not just be called `are_equivalent`?  The `sans_declaration` seems like an implementation detail of the method

lib/zendesk_apps_support/validations/svg.rb
```diff
@@ -0,0 +1,81 @@
...
+            end
+
+            match_pattern = Regexp.new(attr_prop + ":.*?(\;|\z)")
+            clean_attr = attr_value.gsub(match_pattern, '')
+
+            # depending on its contents, strip either the offending property in the attribute, or the entire attribute
+            if clean_attr.empty?
+              filtered_markup_doc.xpath(attr_path).first.remove
+            else
+              filtered_markup_doc.xpath(attr_path).first.value = clean_attr
```
hoangtrvu: /nitpick looks like `filtered_markup_doc.xpath(attr_path)` is repeated many times in this function, can reuse it?

fvzzy: Oh man I didn't see this comment until now. Anyway, I just refactored the whole lot to use Loofah's custom scrubbers - could you take another look at this?

fvzzy: Also I don't think it makes any sense to put this behind Arturo - will discuss in 🍖

lib/zendesk_apps_support/validations/svg.rb
```diff
@@ -7,73 +7,53 @@ module Svg
...
-          markup_doc.root.to_s
-        end
-
-        def are_equivalent(clean_markup, markup)
-          # strip namespaces initially here as it makes searching the document easier,
-          # see: http://www.nokogiri.org/tutorials/searching_a_xml_html_document.html
-          filtered_markup_doc = Nokogiri::XML(markup).remove_namespaces!
+          remove_enable_background = Loofah::Scrubber.new do |node|
+            if node.name == 'svg' and node['style']
+              match_pattern = Regexp.new("enable-background:.*?(\;|\z)")
```
hoangtrvu: `match_pattern` can be defined outside

lib/zendesk_apps_support/validations/svg.rb
```diff
@@ -0,0 +1,57 @@
...
+          end
+
+          empty_malformed_markup = Loofah::Scrubber.new do |node|
+            node.next.remove while node.name == 'svg' && node.next
+          end
+
+          # CRUFT: ignore a (very specific) style attribute which Loofah would otherwise scrub.
+          # This attribute is deprecated (https://www.w3.org/TR/filter-effects/#AccessBackgroundImage)
+          # but is included in many of the test apps used in fixtures for tests in ZAM, ZAT etc.
+          remove_enable_background = Loofah::Scrubber.new do |node|
```
hoangtrvu: all loofah scrubbers can be defined outside of `call` method?

lib/zendesk_apps_support/validations/svg.rb
```diff
@@ -0,0 +1,57 @@
...
+            if node.name == 'svg' && node['style']
+              match_pattern = Regexp.new("enable-background:.*?(\;|\z)")
```
hoangtrvu: `match_pattern` can be defined outside?

lib/zendesk_apps_support/validations/svg.rb
fvzzy: 👍

lib/zendesk_apps_support/validations/svg.rb
fvzzy: 👍

PR #181, opened 2017-09-21
Deprecate standalone chat

xuebingli:
cc @zendesk/wombat @zendesk/vegemite 

🔴 DO NOT MERGE UNTIL ALL REFERENCES HAVE BEEN REMOVED 🔴 
🔴 https://github.com/search?q=org%3Azendesk+standalone_chat&type=Code 🔴 

Standalone chat is dead. Let's put it to rest.

### DIFF
--- a/lib/zendesk_apps_support/location.rb
+++ b/lib/zendesk_apps_support/location.rb
@@ -40,11 +40,9 @@ module ZendeskAppsSupport
                    v2_only: true),
       Location.new(id: 10, orderable: false, name: 'ticket_editor', product_code: Product::SUPPORT.code,
                    v2_only: true),
-      Location.new(id: 11, orderable: false, name: 'nav_bar', product_code: Product::STANDALONE_CHAT.code,
-                   v2_only: false),
+      # nav_bar on standalone_chat (with id 11) has been deprecated
       Location.new(id: 12, orderable: false, name: 'system_top_bar', product_code: Product::SUPPORT.code),
-      Location.new(id: 13, orderable: false, name: 'system_top_bar', product_code: Product::STANDALONE_CHAT.code,
-                   v2_only: false),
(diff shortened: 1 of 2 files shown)


xuebingli:
This PR removes reference to standalone chat in the Chat app https://github.com/zopim/zendesk_app/pull/141

manikj:
As mentioned on https://github.com/zopim/zendesk_app/pull/141,
a few accounts (including Z1) are not yet willing to have the arturo flags `voltron_chat` and `voltron_enabled` turned on, and thus need standalone_chat. Removing this causes the app to break, as Christ

The reason for this is that the voltron world only supports Zopim-Zendesk ticket creation V2 (due to differing authentication mechanisms), while these accounts have a distinct preference to use V1 over V2. There's a plan to unify these feature sets, but no roadmap for it yet

Is this a product push, or just something you ideally want to carry out?
If it's the latter, may I request we hold off on this for a while?

xuebingli:
@manikj TLDR it's for a product push and there's a way to move forward without removing Standalone Chat.
Now the long story. As far as I can remember, this change is needed for [backfilling app translations in staging and production](https://zendesk.atlassian.net/browse/APPS-3574). We are associating all existing app translations with either Zendesk Support or Zendesk Chat, depending on the translation's corresponding app's product. Since Standalone Chat is neither Zendesk Support nor Zendesk Chat, it was causing the backfill to fail. 

I reckon the workaround is to handle Standalone Chat as a special case in the backfill.

@fbvilela May you please look into implementing a special case for Standalone Chat?

manikj:
@xuebingli could we consider standalone chat's translations to be under Zendesk Chat till it goes away?

xuebingli:
@manikj I believe we can. 

@fbvilela note that the backfilling script needs to be updated to handle standalone chat.

xuebingli:
Closing PR due to inactivity.

PR #183, opened 2017-09-25
Run rake i18n:standardize

seangoedecke:
Whoops, I forgot to do this one.

Not sure what consequences this might have - maybe translations are missing for some multi-product translation app errors.


seangoedecke:
@fvzzy

PR #186, opened 2017-10-16
[APPS-3650] Ensure location options url is a string

seangoedecke:
When you pass something that isn't a string as a url in location options (e.g. `true`), it causes an airbrake down the line. This PR fixes that.

https://zendesk.atlassian.net/browse/APPS-3650
https://zendesk.airbrake.io/projects/105401/groups/2060291569700544052?tab=notice-detail&attribute_id=-8115297495119076088&resolved=false&search=undefined%20method

### DIFF
--- a/lib/zendesk_apps_support/manifest/location_options.rb
+++ b/lib/zendesk_apps_support/manifest/location_options.rb
@@ -24,7 +24,6 @@ module ZendeskAppsSupport
         RUBY_TO_JSON.each do |ruby, json|
           instance_variable_set(:"@#{ruby}", options[json])
         end
-        @url = @url.to_s
         @legacy ||= @url == ZendeskAppsSupport::Manifest::LEGACY_URI_STUB
         @auto_load = options.fetch('autoLoad', true)
         @auto_hide ||= false
--- a/lib/zendesk_apps_support/manifest/location_options.rb
+++ b/lib/zendesk_apps_support/manifest/location_options.rb
@@ -24,6 +24,7 @@ module ZendeskAppsSupport
         RUBY_TO_JSON.each do |ruby, json|
           instance_variable_set(:"@#{ruby}", options[json])
         end
+        @url = @url.to_s
         @legacy ||= @url == ZendeskAppsSupport::Manifest::LEGACY_URI_STUB
         @auto_load = options.fetch('autoLoad', true)
         @auto_hide ||= false
--- a/spec/validations/manifest_spec.rb
+++ b/spec/validations/manifest_spec.rb
@@ -256,12 +256,33 @@ describe ZendeskAppsSupport::Validations::Manifest do
     end
   end
 
+  context 'location references a non-string URL' do
+    before do
+      @manifest_hash = {
+        'location' => {
+          'zendesk' => {
+            'ticket_sidebar' => {
+              'url' => true
+            }
+          }
+        }
+      }
+    end
+
+    it 'should have a location error' do
+      require 'byebug'
+      byebug
+      expect(@package).to have_error(/location does not specify a URI/)
+    end
+  end
+
+
   context 'location uri is blank' do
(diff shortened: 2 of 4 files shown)


lib/zendesk_apps_support/manifest/location_options.rb
```diff
@@ -24,6 +24,7 @@ def initialize(location, options)
         RUBY_TO_JSON.each do |ruby, json|
           instance_variable_set(:"@#{ruby}", options[json])
         end
+        @url = @url.to_s
```
lcampbell-zendesk: Rather than coercing to a string should we be throwing an ArgumentError if the url isn't a string?

seangoedecke: If we throw an ArgumentError here, we'd have to catch it and somehow bubble up a validation message (despite this file not being validation-related). IMO it's better to coerce to string and validate elsewhere that the urls look like urls. If I were an app developer I'd appreciate a `"true" is not a valid URL` error message more than a `parameter type invalid` message.

mtsolakiszen: Does the same issue occur if a boolean is passed in other places in the manifest? If so maybe we should add validations that the values are set to the correct types

seangoedecke: The problem here is due to this interesting logic in `location_options.rb`. I don't think it's an issue throughout the manifest.

lcampbell-zendesk: I've had a look at the code and the original airbrake. It looks like the `Manifest` class is used to hold the user input and access it during the validation step. Given that we're pre-validation it seems reasonable for the variable to be set to whatever the user entered, so you're right that we shouldn't be throwing an `ArgumentError`.

Instead I think perhaps we should be doing basic validations on the shape and types of values in the manifest before we do more complex validations that make assumptions about those types. So I think a fix that addresses the root cause is probably adding something like this to `ZendeskAppsSupport::Validations::Manifest#collate_manifest_errors`:

``` ruby
if has_valid_locations?(manifest)
 errors << ban_no_template(manifest) if manifest.iframe_only?
else
  errors << invalid_location_error(manifest)
end
```

Then `ban_no_template`, which expects the location to be a string, is only called if the location is a string. If the user has entered the wrong type then `invalid_location_error` will give them a helpful and specific error message.

lcampbell-zendesk:
``` ruby
            if location_options.url.is_a?(String) && !location_options.url.empty?
```

seangoedecke:
@zendesk/wombat Ready for review!

spec/validations/manifest_spec.rb
```diff
@@ -256,12 +256,33 @@ def create_package(parameter_hash)
...
+              'url' => true
+            }
+          }
+        }
+      }
+    end
+
+    it 'should have a location error' do
+      require 'byebug'
+      byebug
```
lcampbell-zendesk: Should be removed?

seangoedecke: Whoops, yes.

spec/validations/manifest_spec.rb
```diff
@@ -256,12 +256,31 @@ def create_package(parameter_hash)
...
+            'ticket_sidebar' => {
+              'url' => true
+            }
+          }
+        }
+      }
+    end
+
+    it 'should have a location error' do
+      expect(@package).to have_error(/location does not specify a URI/)
```
lcampbell-zendesk: This might be a bit pedantic but isn't it a URL we're looking for not a URI. Or would `urn:isbn:0-486-27557-4` be valid in this instance?

fbvilela:
@seangoedecke ready to merge?

PR #187, opened 2017-10-16
[APPS-3651] Stop zas choking on strings that aren't products

seangoedecke:
Turns out if you specify a location that's just a string, and the string doesn't match a known location name in Support (e.g. "ticket_sidebar"), ZAS will break when it tries to figure out what your products are based on your locations. This PR makes ZAS ignore locations it can't make any sense of.

https://zendesk.airbrake.io/projects/105401/groups/2039767550069022343?tab=notice-detail
https://zendesk.atlassian.net/browse/APPS-3651

### DIFF
--- a/spec/manifest_spec.rb
+++ b/spec/manifest_spec.rb
@@ -229,7 +229,7 @@ describe ZendeskAppsSupport::Manifest do
     context 'when products is a string that is not a location' do
       before { manifest_hash[:location] = 'cool_location' }
 
-      it 'defaults to support without raising an error' do
+      it 'returns no products without raising an error' do
         expect(manifest.products).to eq([])
       end
     end
--- a/spec/manifest_spec.rb
+++ b/spec/manifest_spec.rb
@@ -226,6 +226,22 @@ describe ZendeskAppsSupport::Manifest do
       ])
     end
 
+    context 'when products is a string that is not a location' do
+      before { manifest_hash[:location] = 'cool_location' }
+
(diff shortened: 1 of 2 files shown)


spec/manifest_spec.rb
```diff
@@ -226,6 +226,22 @@
       ])
     end
 
+    context 'when products is a string that is not a location' do
+      before { manifest_hash[:location] = 'cool_location' }
+
+      it 'defaults to support without raising an error' do
```
StaphSynth: Should this be 'defaults to support without raising an error'?

seangoedecke: Nope, thanks

itirkarp:
cc @zendesk/wombat

spec/manifest_spec.rb
```diff
@@ -226,6 +226,22 @@
...
+
+      it 'returns no products without raising an error' do
+        expect(manifest.products).to eq([])
+      end
+    end
+
+    context 'when products is a string that matches a location' do
+      before { manifest_hash[:location] = 'ticket_sidebar' }
+
+      it 'defaults to support without raising an error' do
```
itirkarp: Why would it default to support if it matches a valid location? It would detect the correct product right? The test doesn't change but the name of the test would

seangoedecke: Mmm. Unfortunately, I think it doesn't detect the correct product. If no product is specified ZAS assumes we're in support, then does a search for support locations which match the string in the manifest.

PR #188, opened 2017-10-17
[AF-918] Fix SVG sanitisation in ZAT Validate to work with Zendesk Garden assets

fvzzy:
:v:

/cc @zendesk/vegemite

### Description
Whitelists some elements and properties so the (valid) SVG assets compiled by Zendesk Garden don't trigger a sanitisation rewrite.

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-918
* Z1: https://support.zendesk.com/agent/tickets/3046110

### Risks
* [low] exposes vulnerabilities in additionally allowed elements/properties.


lib/zendesk_apps_support/validations/svg.rb
```diff
@@ -4,33 +4,38 @@
...
       # but is included in many of the test apps used in fixtures for tests in ZAM, ZAT etc.
-      @remove_enable_background = Loofah::Scrubber.new do |node|
-        match_pattern = Regexp.new("enable-background:.*?(\;|\z)")
+      Loofah::HTML5::WhiteList::ACCEPTABLE_CSS_PROPERTIES.add 'enable-background'
+
+      @strip_declaration = Loofah::Scrubber.new do |node|
+        node.remove if node.name == 'xml' && node.children.empty?
+      end
+
+      @strip_spaces_between_css_attrs = Loofah::Scrubber.new do |node|
```
danielbreves: Why do we do this?

lib/zendesk_apps_support/validations/svg.rb
```diff
@@ -4,33 +4,38 @@
...
+        match_pattern = Regexp.new(/\;\s+/)
         if node.name == 'svg' && node['style']
-          node['style'] = node['style'].gsub(match_pattern, '')
-          node.attributes['style'].remove if node['style'].empty?
+          node['style'] = node['style'].gsub(match_pattern, ';')
         end
       end
 
+      @empty_malformed_markup = Loofah::Scrubber.new do |node|
+        node.next.remove while node.name == 'svg' && node.next
```
danielbreves: Is this because we can only support one svg node?

spec/validations/svg_spec.rb
```diff
@@ -2,11 +2,19 @@
 require 'spec_helper'
 
 describe ZendeskAppsSupport::Validations::Svg do
-  let(:clean_markup) do
+  clean_markup = [
```
danielbreves: Could we define this within the 'clean markup' context for consistency?

spec/validations/svg_spec.rb
```diff
@@ -100,7 +110,7 @@
       let(:markup) { markup }
       it 'sanitises questionable markup and notifies the user that the offending svgs were modified' do
         errors = ZendeskAppsSupport::Validations::Svg.call(package)
-        expect(IO).to have_received(:write).with(svg.relative_path, clean_markup)
+        expect(IO).to have_received(:write).with(svg.relative_path, clean_markup[0])
```
danielbreves: It doesn't seem clear that the first `clean_markup` is a sanitised version of these. Could we redefine it here to be explicit?

lib/zendesk_apps_support/validations/svg.rb
fvzzy: Loofah condenses these spaces in it's optimisation of the markup, so stripping them before sanitising ensures we don't unnecessarily trigger a rewrite.* 

*(This is if you're referring to the `@strip_spaces_between_css_attrs` - I think you are, but its hard to tell)

lib/zendesk_apps_support/validations/svg.rb
fvzzy: Yep, sort of. I think it was also so we didn't have subsequent tags following/outside the main <svg>. I do remember adding this to make the tests at https://github.com/zendesk/zendesk_apps_support/blob/cd87351ed0bff97347c655e59a3e512b5ba572b6/spec/validations/svg_spec.rb#L120 make sense

lib/zendesk_apps_support/validations/svg.rb
danielbreves: Could you add a comment please?

fvzzy:
Example steps to QA:
1. Checkout https://github.com/zopim/chat-dynamic-content-app
2. `webpack` to build assets into `dist`
3. Inside local ZAT, temporarily point ZAS dependency to this branch
4. Run `zat validate` using local ZAT inside `chat-dynamic-content-app`
5. Validation should pass without triggering a rewrite on `index.svg`

PR #190, opened 2017-10-18
[APPS-3602] Manifest validation.Return error instead of exploding

fbvilela:
@zendesk/wombat @zendesk/vegemite 

### Description
returns an error which is more user friendly. 
screen shot 2017-10-18 at 5 02 46 pm

Prior to this you would see a `something went wrong message` due to an exception thrown at the worker level. 

This also reduces airbrakes.

### Risks
* low - breaks manifest validation

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -22,6 +22,8 @@ en:
             no_app_js_required: Do not set your app to requirements only if you need
               app.js
             manifest_not_json: manifest is not proper JSON. %{errors}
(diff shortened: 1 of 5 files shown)


danielbreves:
@fbvilela @mtsolakiszen I'm getting `uninitialized constant ZendeskAppsSupport::Manifest::OverrideError (NameError)` when trying to `zat package` I think we may need to `autoload OverrideError` [here](https://github.com/zendesk/zendesk_apps_support/blob/master/lib/zendesk_apps_support.rb#L33-L37).

PR #192, opened 2017-11-03
Fix path to SVG file for writing sanitised markup

fvzzy:
:v: 🐨 

### Description
Points to an SVG path that actually exists for the ZAM worker to (re)write to.

### References
* Slack: https://zendesk.slack.com/archives/C122US89L/p1509644506000190

### Risks
* [low] safe SVGs are rewritten and don't display as initially intended


fbvilela:
FYI - pushed a commit with rubucop fixes

PR #194, opened 2017-11-29
[AF-952] Warn when bitmaps are included in SVGs and replace with a placeholder…

fvzzy:
:v:

/cc @zendesk/vegemite

### Description
As per the developer docs, [SVG icons submitted for a top or nav bar app should not include bitmaps ](https://developer.zendesk.com/apps/docs/publish/create_assets#dont-include-colors-gradients-shadows-or-bitmaps).

This PR warns the developer when there is a bitmap/raster image embedded within their SVG icon, and replaces the icon with a placeholder image.

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-952

### Risks
* Low - causes unexpected overwrites for valid SVG icons to default placeholder images.


lib/zendesk_apps_support/validations/svg.rb
```diff
@@ -31,27 +39,43 @@ module Svg
...
-
-            clean_markup = Loofah.xml_fragment(markup)
-                                 .scrub!(:prune)
-                                 .scrub!(@empty_malformed_markup)
-                                 .to_xml
-
-            next if clean_markup == markup
-            begin
+            if contains_embedded_bitmap?(svg.read)
+              rewrite_svg(svg, PLACEHOLDER_SVG_MARKUP, package, errors)
```
hoangtrvu: should we only clean the `<image>` part and still continue the next scrub steps?

lib/zendesk_apps_support/validations/svg.rb
```diff
@@ -4,6 +4,14 @@
 module ZendeskAppsSupport
   module Validations
     module Svg
+      PLACEHOLDER_SVG_MARKUP = %(<?xml version="1.0" encoding="utf-8"?><svg version="1.1" id="Layer_1" \
```
fvzzy: Take this out

lib/zendesk_apps_support/validations/svg.rb
```diff
@@ -31,27 +33,43 @@ module Svg
...
+
+        def rewrite_svg(svg, new_markup, package, errors)
+          warning_string = if contains_embedded_bitmap?(svg.read)
+                             'txt.apps.admin.warning.app_build.bitmap_in_svg'
+                           else
+                             'txt.apps.admin.warning.app_build.sanitised_svg'
+                           end
+
+          package.warnings << I18n.t(warning_string, svg: svg.relative_path)
+          compressed_new_markup = new_markup.tr("\n", '').squeeze(' ').gsub(/\>\s+\</, '><')
```
ocke: why do we need to parse this new_markup, isn't it coming from 'default_app_logo.svg', so it will be correct?

fvzzy: I chose to store the icon uncompressed in the repo and looking like the markup that comes from Garden - I thought it was a little more accessible that way. Passing it back through here so that on the server it's stored as the stripped down/compressed version.

fvzzy:
@zendesk/vegemite steps to QA:

1) locally, point ZATs ZAS dependency to this branch (add `:git => 'https://github.com/zendesk/zendesk_apps_support.git', :branch => 'farhan/af-952-catch-embedded-bitmaps-in-svgs'` to the Gemfile)
2) `zat validate` against an app that has a bitmap included in one of the icons, e.g. https://support.zendesk.com/attachments/token/mHCB57utxs4jPGEsuegrquLmf/?name=546517.zip
3) the icon should be replaced with a placeholder and an appropriate warning logged out

token-cjg:
:+1:

PR #197, opened 2017-12-11
[AF-910] Add 'collapsible' to app location properties

fvzzy:
:v:

/cc @zendesk/vegemite

### Description
Adds a `collapsible` flag to the app location properties. For consumption in AF-910 (will link)

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-910

### Risks
* None/Low - causes issues in consumers utilising all of the properties of the `Location` class

### DIFF
--- a/lib/zendesk_apps_support/location.rb
+++ b/lib/zendesk_apps_support/location.rb
@@ -2,7 +2,7 @@
 module ZendeskAppsSupport
   class Location
     extend ZendeskAppsSupport::Finders
-    attr_reader :id, :name, :orderable, :product_code, :v2_only
+    attr_reader :id, :name, :orderable, :collapsible, :product_code, :v2_only
 
     def self.unique_ids
       @ids ||= Set.new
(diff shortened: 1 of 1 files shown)


lib/zendesk_apps_support/location.rb
```diff
@@ -14,6 +14,7 @@ def initialize(attrs)
       Location.unique_ids.add @id
       @name = attrs.fetch(:name)
       @orderable = attrs.fetch(:orderable)
+      @collapsible = attrs.fetch(:collapsible)
```
svizzari: I wonder if we could have implemented collapsible similar to `v2_only`, defaulting to `false`, so it would only need to be declared for a Location where it was `true`?

danielbreves: We can make it default to either value in the constructor, no?

fvzzy: Yeah I did think about this. For consistency I might update this so it's set the same way as `v2_only`

svizzari: > We can make it default to either value in the constructor, no?

I described as defaulting to `false`, cos' it looks like (similar to `v2_only`), the majority case is `false`, meaning less lines of code to declare `true` where needed

danielbreves: Sure, I was thinking the same could be applied to the `orderable` flag.

lib/zendesk_apps_support/location.rb
```diff
@@ -30,22 +31,24 @@ def self.all
     LOCATIONS_AVAILABLE = [
```
danielbreves: This is getting quite difficult to visually parse. Could we setup some defaults on the constructor to reduce the redundancy?

PR #198, opened 2018-02-01
Add collapsible attribute to installations

Mehul-Porwal:
Adding a new attribute to the installations class, this is to be used in lotus to check if a particular app instance is collapsible or not.

Required for: https://github.com/zendesk/zendesk_app_market/pull/2888

@zendesk/vegemite

### DIFF
--- a/lib/zendesk_apps_support/installation.rb
+++ b/lib/zendesk_apps_support/installation.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 module ZendeskAppsSupport
   class Installation
-    attr_accessor :id, :app_id, :app_name, :requirements, :settings, :enabled, :updated_at, :created_at, :collapsible
+    attr_accessor :id, :app_id, :app_name, :requirements, :settings, :enabled, :collapsible, :updated_at, :created_at
 
     def initialize(options)
       options.each do |k, v|
--- a/lib/zendesk_apps_support/installation.rb
+++ b/lib/zendesk_apps_support/installation.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 module ZendeskAppsSupport
   class Installation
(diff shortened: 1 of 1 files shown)


lib/zendesk_apps_support/installation.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 module ZendeskAppsSupport
   class Installation
-    attr_accessor :id, :app_id, :app_name, :requirements, :settings, :enabled, :updated_at, :created_at
+    attr_accessor :id, :app_id, :app_name, :requirements, :settings, :enabled, :updated_at, :created_at, :collapsible
```
fvzzy: Hmm I'm not sure this makes sense. `collapsible` was designed to be an attribute that defines whether or not a location is collapsible at all (i.e. only sidebar apps are collapsible, so only those ones get the lock displayed in the manage page).

Is this new attribute persisting the state of the the app being expanded/collapsed? Perhaps just `collapsed` or `visible` then?

Mehul-Porwal: This is not persisting the state of the app... but lotus needs to know whether a app is `collapsible` or not. Based on that it will either have the app expanded and no option to collapse(lock icon) to have the chevron which can be expanded and collapsed. Does this make sense?

Mehul-Porwal: From your PR(https://github.com/zendesk/zendesk_app_market/pull/2834/files) I see you have used collapsible as an attribute to 2 different objects:
`location.collapsible`
`installation.collapsible`

This PR is just using `installation.collapsible`, `location.collapsible` is specific to ZAM, not required for lotus.

fvzzy: Yep, makes sense 👍 Does this need to be read in at the installation level though?

Also nitpick but can `:updated_at` and `:created_at` be the last attributes in that list also? Seems like they usually are

PR #199, opened 2018-02-02
[CACT-354] Validates translation file when parameters are included

fbvilela:
🐻 

### Description

 Adds a new validation to each translation file, asserting it has a `label` key on each node under `parameters` . See https://developer.zendesk.com/apps/docs/agent/i18n for example.

### References
* JIRA: https://zendesk.atlassian.net/browse/CACT-354

### Risks
* [low] breaks translation validation

### DIFF
--- a/config/locales/translations/zendesk_apps_support.yml
+++ b/config/locales/translations/zendesk_apps_support.yml
@@ -255,6 +255,11 @@ parts:
       title: "App builder job: required key missing from translation file"
       value: "Missing required key from %{file}: %{missing_key}"
   - translation:
+      key: "txt.apps.admin.error.app_build.translation.missing_required_key_on_leaf"
+      title: "App builder job: required key missing from translation file."
+      screenshot: "https://zendesk.box.com/s/m6yr2j0vai0a7qkimj2gcqiy61jznmn4"
+      value: "Missing required key %{missing_key} on leaf %{leaf} from %{file}"
+  - translation:
       key: "txt.apps.admin.error.app_build.translation.missing_required_key_for_product"
       title: "App builder job: required key missing from translation file for a specific product"
       value: "Missing required key from %{file} for %{product}: %{missing_key}"
--- a/lib/zendesk_apps_support/validations/translations.rb
+++ b/lib/zendesk_apps_support/validations/translations.rb
@@ -16,7 +16,7 @@ module ZendeskAppsSupport
           package.files.each_with_object([]) do |file, errors|
             path_match = TRANSLATIONS_PATH.match(file.relative_path)
             next unless path_match
-            errors << locale_error(file, path_match[1]) << json_error(file)
+            errors << locale_error(file, path_match[1]) << json_error(file) << format_error(file)
             next unless errors.compact.empty?
             errors.push(*validate_marketplace_content(file, package)) if file.relative_path == 'translations/en.json'
           end.compact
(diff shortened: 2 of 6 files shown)


fbvilela:
Couldn't use `Hash#dig` because it belongs to newer ruby versions.

lib/zendesk_apps_support/validations/translations.rb
```diff
@@ -29,9 +29,23 @@ def locale_error(file, locale)
           ValidationError.new('translation.invalid_locale', file: file.relative_path)
         end
 
+        def format_error(file)
+          with_valid_json(file) do |json|
+            if json['app'] && json['app']['parameters']
```
waferbaby: Could you use `dig` here?

lib/zendesk_apps_support/validations/translations.rb
```diff
@@ -41,6 +55,13 @@ def json_error(file)
                 ValidationError.new('translation.invalid_format', field: e.message)
               end
             end
+          end
+        end
+
+        def with_valid_json(file)
```
seangoedecke: Nice

lib/zendesk_apps_support/validations/translations.rb
```diff
@@ -29,9 +29,23 @@ def locale_error(file, locale)
...
         end
 
+        def format_error(file)
+          with_valid_json(file) do |json|
+            if json['app'] && json['app']['parameters']
+              parameters_node = json['app']['parameters']
+              parameters_node.keys.map do |param|
+                unless parameters_node[param].is_a?(Hash) && parameters_node[param].keys.include?('label')
+                  return ValidationError.new('translation.missing_required_key',
+                                             file: file.relative_path, missing_key: "#{param} must have 'label'")
```
seangoedecke: Will this work when this error message is translated? I reckon we'll probably need a new translation instead (e.g. `missing_label_key_for_parameter`) so that "must have 'label'" can actually get translated

fbvilela: Oh yeah, good point. I wasn't 100% sure if we do have translations, other than English, for those keys. I could simplify things with the error message to make use of the same key. What do you think?

seangoedecke: Probably work asking Phil how good the error message needs to be for this. If it's a very rare problem then I think it's OK to use the current error message and simplify it. But if people are likely to get this wrong a lot then we should probably add another key :man_shrugging:

fbvilela:
cc @zendesk/i18n

fbvilela:
@zendesk/vegemite how do you make the tests go green when adding a new translation?

danielbreves:
@fbvilela have you run [`bundle exec rake i18n:standardize`](https://github.com/zendesk/zendesk_apps_support/blob/f4a6eb2a77f795bb661d41c1300038449f60f99d/config/locales/translations/zendesk_apps_support.yml#L1-L3)?

config/locales/translations/zendesk_apps_support.yml
```diff
@@ -254,6 +254,10 @@ parts:
       key: "txt.apps.admin.error.app_build.translation.missing_required_key"
       title: "App builder job: required key missing from translation file"
       value: "Missing required key from %{file}: %{missing_key}"
+  - translation:
+      key: "txt.apps.admin.error.app_build.translation.missing_required_key_on_leaf"
+      title: "App builder job: required key missing from translation file"
+      value: "Missing required key %{missing_key} on leaf %{leaf} from %{file}"
```
mariasica: @hiroiguchi please add info/screenshots or translations will come back funny for sure ;-)

mariasica: guys it is better to create separate PR for strings only so they can be merged quicker than the code (since translation take time) - best practice only. Thanks!

fbvilela: @hiroiguchi I added a screenshot. would that be enough?

config/locales/translations/zendesk_apps_support.yml
```diff
@@ -254,6 +254,10 @@ parts:
       key: "txt.apps.admin.error.app_build.translation.missing_required_key"
       title: "App builder job: required key missing from translation file"
       value: "Missing required key from %{file}: %{missing_key}"
+  - translation:
+      key: "txt.apps.admin.error.app_build.translation.missing_required_key_on_leaf"
+      title: "App builder job: required key missing from translation file. example: https://zendesk.box.com/s/m6yr2j0vai0a7qkimj2gcqiy61jznmn4"
+      value: "Missing required key %{missing_key} on leaf %{leaf} from %{file}"
```
mariasica: for the string. Always better to create separate pr for strings only @hiroiguchi

hiroiguchi: @mariasica Noted. Thank you for your advice.

@fbvilela Thank you for uploading an image. can you add `screenshot: "https://zendesk.app.box.com/s/m6yr2j0vai0a7qkimj2gcqiy61jznmn4"` between title and value please?

fbvilela: all done

hiroiguchi:
@fbvilela Thank you 👌

PR #201, opened 2018-02-13
don't run translation code for non-v2 apps.

ocke:
Easy peasy, lemon skeasy PR for not error when we don't have translations in V2.


lib/zendesk_apps_support/package.rb
```diff
@@ -130,7 +130,7 @@ def compile(options)
         location_icons: location_icons,
         app_class_name: app_class_name,
         author: manifest.author,
-        translations: runtime_translations(translations_for(locale)),
+        translations: manifest.iframe_only? ? nil : runtime_translations(translations_for(locale)),
```
ocke: `iframe_only?` means `is a v2 app?`

PR #202, opened 2018-02-14
Valid {{setting.}} url

ocke:
/cc @zendesk/vegemite 

If we have a setting we can't check the url for validity, because we don't know how the url is setup.

This will validate any url setting that has a `{{setting.`` in it.
```
url: "{{setting.url}}"
url: "http://{{setting.domain}}.my_server.com/iframe.html"
url: "/assets/{{setting.page}}.html"
```

### DIFF
--- a/lib/zendesk_apps_support/validations/manifest.rb
+++ b/lib/zendesk_apps_support/validations/manifest.rb
@@ -231,7 +231,7 @@ module ZendeskAppsSupport
           errors = []
           package.manifest.location_options.each do |location_options|
             if location_options.url.is_a?(String) && !location_options.url.empty?
-              errors << invalid_location_uri_error(package, location_options.url)
+              errors << invalid_location_uri_error(package, location_options)
             elsif location_options.auto_load?
               errors << ValidationError.new(:blank_location_uri, location: location_options.location.name)
             end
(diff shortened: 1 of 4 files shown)


lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -272,6 +272,7 @@ def invalid_v1_location(package)
 
         def invalid_location_uri_error(package, path)
           return nil if path == ZendeskAppsSupport::Manifest::LEGACY_URI_STUB
+          return nil if path.index('{{setting.') != nil
```
vaevictis: Or if you want to write it shorter:
`return nil if path == ZendeskAppsSupport::Manifest::LEGACY_URI_STUB || path.index('{{setting.') != nil`

danielbreves:
Can we validate that `signed` and/or `signedUrls` is `false`?

ocke:
what is `signed_urls` if have never heard of that before.

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -270,8 +270,10 @@ def invalid_v1_location(package)
           end
         end
 
-        def invalid_location_uri_error(package, path)
-          return nil if path == ZendeskAppsSupport::Manifest::LEGACY_URI_STUB
+        def invalid_location_uri_error(package, location_options)
+          path = location_options.url
+          return nil if path == ZendeskAppsSupport::Manifest::LEGACY_URI_STUB ||
+                        (!location_options.signed && !path.index('{{setting.').nil?)
```
danielbreves: Could you add tests for these please?

danielbreves: Could probably use [`include?`](http://ruby-doc.org/core-1.9.3/String.html#method-i-include-3F) to check the substring.

danielbreves:
[`signedUrls`](https://developer.zendesk.com/apps/docs/developer-guide/manifest#signedurls) makes all locations signed.

lib/zendesk_apps_support/validations/manifest.rb
danielbreves: Could we return a more descriptive error message if the url is signed and has a setting placeholder?

PR #204, opened 2018-03-09
fix error message

ocke:
/cc @zendesk/vegemite 

Fixed the error message. Also passed the path param!


hiroiguchi:
For string, 👌

hiroiguchi:
cc @zendesk/i18n

PR #207, opened 2018-04-18
Tell installed.js if the app is private

seangoedecke:
This is a proof of concept.


seangoedecke:
We're at 100% and the traffic is manageable. I'm closing this out since we don't need it.

PR #208, opened 2018-04-20
[LOCAL-44] Strings: reverse sweep (auto)

string-sweep:
NOTE: String Sweep will be suspended for this repo until this PR is merged or closed!

/cc @zendesk/i18n @zendesk/i18n-dev
### Risks
Low. English edits and title improvements.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -19,8 +19,6 @@ en:
               only
             no_framework_version_required: Framework versions can't be set when you
               specify requirements only
-            no_app_js_required: Do not set your app to requirements only if you need
-              app.js
             manifest_not_json: manifest is not proper JSON. %{errors}
             duplicate_manifest_keys: Duplicate keys are specified in the manifest.json
               file. %{errors}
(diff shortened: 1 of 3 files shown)


coreysf:
@zendesk/vegemite - Please check out the failing test here. This will need to be merged before the repo will be included in the next sweep. Thanks!

danielbreves:
@coreysf tl;dr it's fixed now.

ZAS requires a rake command (rake i18n:standardize) to be run after updating translations to make them compatible with the i18n gem (we can't use the Zendesk one because this is an open-source project). We have a test that makes sure the generated file was updated after the zendesk one. Now because @string-sweep only changed the zendesk one it failed. I had to make a manual change to the generated one to fix it.

Any way we can make @string-sweep do this automatically?

dadah89:
Took a look with the I18n-dev team. Running the rake task won't be possible (it's not feasible to clone repos when sweeping files), so was looking at ways of moving the logic somewhere else (without creating one-off solutions and making the string sweep more complex). But looking at the rake task, it basically copies the data to a new file, which isn't good at all, it just makes things more complex because now you need to sync both files.

I'd say the correct solution, Instead of maintaining the rake task plus keeping the file in sync, would be to just implement your own I18n backend (http://guides.rubyonrails.org/i18n.html#customize-your-i18n-setup) that knows how to consume `zendesk_apps_support.yml` (it's just a yaml file anyways, it isn't too complicated). Also an option, If your team would like, we can work together to extract that backend from our private i18n gem, so you can use it here directly. In the meantime, we'll need to ask that your team fixes the build for reverse sweep PRs, we deal with a lot of repos per week, even checking branches in each repo to fix the build (whether it be tests or running rake tasks) isn't feasible for us =\

If you'd like to discuss more, feel free to ping me or the team.

danielbreves:
Hi @dadah89, thanks for looking into this! I'll look into these options. In the meantime we'll make sure to fix the build for reverse sweep PRs. Apologies for the trouble.

dadah89:
@danielbreves we're constantly thinking of ways to have less manual steps for us and for the dev teams, knowing better about the different use cases definitely help us, so send any feedback you have! We'd be pretty happy to work together on a fix too, don't be afraid of reaching out.

PR #210, opened 2018-05-22
Handle nil framework version

mtsolakiszen:

RubyGems v2.7 introduces a change to the `Gem::Version` class which throws an error if an instance is initialised with `nil`. 

https://github.com/rubygems/rubygems/blob/2.7/lib/rubygems/version.rb#L173

Requirements only apps cannot have a specified framework version and were previously having the `iframe_only?` method return `false` but with the latest version of RubyGems this throws an error.


token-cjg:
:+1:

PR #212, opened 2018-06-28
[AF-1050] [AF-1049] Permit and validate custom resource schemas defined in requirements.json

fvzzy:
:v:

/cc @zendesk/vegemite @zendesk/i18n 	

### Description
1. Allows `custom_resources_schema` to be defined within `requirements.json`, and
1. Does some basic client-side validation to ensure that the schema contains only valid and all required keys for the custom resources `/private/jobs` endpoint.

### References
* https://zendesk.atlassian.net/browse/AF-1050
* https://zendesk.atlassian.net/browse/AF-1049

### Risks
* Low - just passes validation on a new resource type when defined in an app.

PR #214, opened 2018-07-19
Permit and validate custom resource schemas defined in requirements.json

fvzzy:
Restores https://github.com/zendesk/zendesk_apps_support/pull/212 with syntax fixed for Ruby 2.1

### DIFF
--- a/lib/zendesk_apps_support/validations/requirements.rb
+++ b/lib/zendesk_apps_support/validations/requirements.rb
@@ -5,7 +5,7 @@ module ZendeskAppsSupport
     module Requirements
       MAX_REQUIREMENTS = 5000
 
-      class <<self
+      class << self
         def call(package)
           if package.manifest.requirements_only? && !package.has_requirements?
             return [ValidationError.new(:missing_requirements)]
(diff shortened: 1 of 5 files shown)


lib/zendesk_apps_support/validations/requirements.rb
```diff
@@ -76,19 +81,40 @@ def invalid_channel_integrations(requirements)
           return unless channel_integrations
           [].tap do |errors|
             if channel_integrations.size > 1
-              errors << ValidationError.new(:multiple_channel_integrations, count: channel_integrations.size)
```
ocke: We don't need count anymore? not really sure what count is used for.

fvzzy: Yeah good question. As far as I can tell, `count` gets used in the translations when they are formatted for example like:
```
- translation:
      key: "txt.apps.admin.error.app_build.duplicate_requirements.one"
      title: "App builder job: requirements file contains duplicate key error"
      value: "requirements.json contains a duplicate key: %{duplicate_keys}"
  - translation:
      key: "txt.apps.admin.error.app_build.duplicate_requirements.other"
      title: "App builder job: requirements file contains duplicate keys error"
      value: "requirements.json contains duplicate keys: %{duplicate_keys}"
```

Note the `.one` and `.other` suffixes. The `missing_required_fields` doesn't handle a case like this though, i.e:

```
- translation:
      key: "txt.apps.admin.error.app_build.missing_required_fields"
      title: "App builder job: required key missing in requirements, e.g. \"title\" is required in \"my_custom_email_target\""
      value: "Missing required fields in requirements.json: \"%{field}\" is required in \"%{identifier}\""
```

...so I don't _think_ it needs this param passed.

PR #215, opened 2018-07-24
[WIP] Wrap global jQeury in AppScope

Mehul-Porwal:
⚠️ DO NOT MERGE

@zendesk/vegemite Just putting this up here for early review.


lib/zendesk_apps_support/assets/installed.js.erb
```diff
@@ -1,4 +1,19 @@
 (function() {
+  var self = this;
+  var reporter = new window.ZendeskReporter();
```
danielbreves: Will probably need a guard here, I don't think `ZendeskReporter` is available in Chat.

lib/zendesk_apps_support/assets/installed.js.erb
```diff
@@ -1,4 +1,19 @@
...
+  function wrapped$() {
+    reporter.increment(
+      'app_framework.app_scope.violated',
+      1,
+      ['origin:' + self.location.hostname]
+    );
+    return self.$.apply(null, arguments);
+  }
+
+  var wrappedWindow = this.$.extend(true, {}, self, {$: wrapped$});
```
Mehul-Porwal: NOTE: Will need a guard here too, Chat does not use jQuery.

danielbreves: jQuery is a hard dependency of ZAF, so we include it in chat here: https://github.com/zopim/meshim-frontend/blob/master/src/meshim/dashboard/controllers/ZendeskApps.js#L153.

Mehul-Porwal: Right, completely forgot about this!

seangoedecke:
What's `window.ZendeskReporter`?

Mehul-Porwal:
window.ZendeskReport is an instance of https://github.com/zendesk/zendesk_http_logger/blob/master/lib/zendesk_reporter.js

It is used to send data to datadog.

fvzzy:
Nit: following our git talk yesterday, could you fix the typo in your commit message please?

(Should be: _Wrap global **jQuery** in AppScope_)

lib/zendesk_apps_support/installed.rb
```diff
@@ -17,7 +17,8 @@ def compile(options = {})
         appsjs: @appsjs,
         installations: @installations,
         installation_orders: options.fetch(:installation_orders, {}),
-        rollbar_zaf_access_token: options.fetch(:rollbar_zaf_access_token, '')
+        rollbar_zaf_access_token: options.fetch(:rollbar_zaf_access_token, ''),
+        zaf_naughty_v1_logging: options.fetch(:zaf_naughty_v1_logging, '')
```
fvzzy: I think our Arturos are normally prefixed with `apps_`. Should that be the case here?

lib/zendesk_apps_support/assets/installed.js.erb
```diff
@@ -1,6 +1,28 @@
...
+  }
+
+  var wrappedWindow = this.$.extend(true, {}, self, { $: wrapped$ });
+  wrappedWindow.window = wrappedWindow;
+  wrappedWindow.top = wrappedWindow;
+
   <% appsjs.each do |appjs| %>
-    <%= appjs %>
+    <% if zaf_naughty_v1_logging %>
+      <%= appjs.gsub(
```
fvzzy: Is it possible to (easily) test .erb templates? If so, it'd be nice to see a simple test for this regex

lib/zendesk_apps_support/installed.rb
danielbreves: Yes, good catch

lib/zendesk_apps_support/assets/installed.js.erb
Mehul-Porwal: 👍 I'll add a test case for this.

lib/zendesk_apps_support/assets/installed.js.erb
```diff
@@ -1,6 +1,28 @@
...
+  var wrappedWindow = this.$.extend(true, {}, self, { $: wrapped$ });
+  wrappedWindow.window = wrappedWindow;
+  wrappedWindow.top = wrappedWindow;
+
   <% appsjs.each do |appjs| %>
-    <%= appjs %>
+    <% if apps_zaf_naughty_v1_logging %>
+      <%= appjs.gsub(
+        /}\(\)\)\;*\s*;*\s*}\s*var\s*app\s*=\s*ZendeskApps\.defineApp\(source\)/mx,
+        "}.bind(wrappedWindow)());\n;\n}\n\nvar app = ZendeskApps.defineApp(source)"
```
danielbreves: I wonder if we could format this properly and convert the whitespace and line breaks programatically to make it easier to read?

Mehul-Porwal: I'm not sure how to do that. In the replaced text, we can just remove the whitespaces, if that will help with readability.

Mehul-Porwal: Or I can add a comment block that tell you what is the regex replacing and with what in a nice formatted way.

lib/zendesk_apps_support/assets/installed.js.erb
```diff
@@ -1,6 +1,28 @@
...
+  function wrapped$() {
+    reporter.increment && reporter.increment(
+      'app_framework.app_scope.violated',
+      1,
+      ['origin:' + self.location.hostname]
+    );
+    return self.$.apply(null, arguments);
+  }
+
+  var wrappedWindow = this.$.extend(true, {}, self, { $: wrapped$ });
```
hoangtrvu: I think you should leave a comment here briefly explaining the purpose of these 3 lines (L13-15)

lib/zendesk_apps_support/assets/installed.js.erb
hoangtrvu: A comment here would be good 👍

spec/fixtures/installed_no_logging.js
```diff
@@ -0,0 +1,80 @@
...
+  var wrappedWindow = this.$.extend(true, {}, self, { $: wrapped$ });
+  wrappedWindow.window = wrappedWindow;
+  wrappedWindow.top = wrappedWindow;
+
+  with( ZendeskApps.AppScope.create() ) {
+    require.modules = {
+        "a.js": function(exports, require, module) {
+          var a = {
+    name: 'This is A'
+  };
```
hoangtrvu: The alignment in this file looks broken, same for other fixture files.

Mehul-Porwal: I've just copy pasted it from another fixture `legacy_app_en.js`, and the broken alignment of the other code actually matches the way `installed.js` looks in production.

Mehul-Porwal:
@danielbreves @hoangtrvu I added those comments, do you mind reviewing it once again?

PR #216, opened 2018-08-23
[AF-1128] [Patch] Make shallow copy of window instead of deep

Mehul-Porwal:
Making a shallow copy of window as window object can have references to other iframes, and the browser will not allow us deep copy an iframe from a different origin.

Error in browser without this fix:
image

@zendesk/vegemite

### DIFF
--- a/lib/zendesk_apps_support/assets/installed.js.erb
+++ b/lib/zendesk_apps_support/assets/installed.js.erb
@@ -16,7 +16,7 @@
 
   // Adding the wrapped version of jQeury to a clone of window object, which
   // will be used to bind the app's scope.
-  var wrappedWindow = this.$.extend(true, {}, self, { $: wrapped$ });
+  var wrappedWindow = this.$.extend({}, self, { $: wrapped$ });
   wrappedWindow.window = wrappedWindow;
   wrappedWindow.top = wrappedWindow;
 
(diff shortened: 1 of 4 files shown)


keitheous:
👍 true~

lib/zendesk_apps_support/assets/installed.js.erb
```diff
@@ -1,33 +1,51 @@
...
-  wrappedWindow.top = wrappedWindow;
+  <% if apps_zaf_naughty_v1_logging %>
+    var self = this;
+    var reporter = window.ZendeskReporter ? new window.ZendeskReporter() : {};
+
+    function getWrappedWindow(appName) {
+      // This is just defining a wrapped version of jQuery which simply makes a
+      // pass through call to global window.jQuery and adds logging information
+      // about the origin of the page to Datadog.
+      function wrapped$(appName) {
```
hoangtrvu: `appName` should not be here

lib/zendesk_apps_support/assets/installed.js.erb
```diff
@@ -1,33 +1,51 @@
...
+        var tags = ['origin:' + self.location.hostname];
+        if (this.__app_name) {
```
hoangtrvu: How does `this.__app_name` work here? Can we just use `appName` directly?

Mehul-Porwal: Yes, this is leftover from some previous code change, forgot to remove it.

lib/zendesk_apps_support/assets/installed.js.erb
Mehul-Porwal: 👍

lib/zendesk_apps_support/assets/installed.js.erb
Mehul-Porwal: Removed it! 👍

PR #217, opened 2018-11-16
[LAB] Add validations for screenshots and deprecated asset references

fvzzy:
:v:

/cc @zendesk/vegemite @Eshen @zendesk/i18n 

### Description
Adds some basic validations to help:
1) catch some of the basic screenshot issues we repeatedly see when reviewing app submissions (i.e. too many, wrong dimensions), and
2) deprecate usage of `assets.zendesk.com` for both Garden (in preference for jsDelivr) and SDK (moved to S3 static assets bucket).

### Risks
* Low - causes confusion for developers when their apps are rejected during upload.


lib/zendesk_apps_support/validations/libraries.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module ZendeskAppsSupport
+  module Validations
+    module Libraries
+      GARDEN_CDN_URL = /<link\s.*href="(https?:\/\/\S*assets\.zendesk\.com\S*zendesk_garden\.css).*>/
+      LEGACY_SDK_URL = /<script\s.*src="(https?:\/\/\S*assets\.zendesk\.com\S*\/((\d(\.\d+){1,2})|latest)\/zaf_sdk\.js).*>/
```
danielbreves: We should probably update the [docs](https://github.com/zendesk/zaf_docs/blob/16071338868b19975d37307c52d0654b24c06757/doc/v2/dev_guide/getting_started.md#L25) first 😅

fvzzy: Yeah that's fine, and can do. I just added this while I was adding the Garden thing since it seemed I could collate them both underneath the `Validations::Libraries` banner

fvzzy: Not sure how solid these regex are - happy for suggestions on both

lib/zendesk_apps_support/package.rb
```diff
@@ -263,6 +272,10 @@ def has_valid_manifest?(errors)
       has_manifest? && errors.flatten.empty?
     end
 
+    def requirements_marketing_iframe_only_app?(manifest)
+      manifest.requirements_only? || manifest.marketing_only? || manifest.iframe_only?
+    end
```
keitheous: 👍

lib/zendesk_apps_support/validations/screenshots.rb
```diff
@@ -0,0 +1,36 @@
...
+
+      class << self
+        def call(package)
+          errors = []
+          screenshot_paths = Dir["#{package.path_to('assets')}/*"].grep(/screenshot-\d.png$/)
+          file_count = screenshot_paths.count
+          if file_count > MAX_FILES
+            errors << ValidationError.new('screenshots.excessive_files',
+                                          file_count: file_count,
+                                          supported_count: MAX_FILES)
```
keitheous: This is not a blocker. Personally I would prefer to [align the parameters of a method call](https://github.com/rubocop-hq/ruby-style-guide#no-double-indent) if they span more than one line.

keitheous:
NICEEEE!

lib/zendesk_apps_support/package.rb
fvzzy: Personally I find this a bit unnecessary but included it as otherwise Rubocop complained the complexity of the `Package` class was too high. Good suggestion on making these private methods

lib/zendesk_apps_support/validations/screenshots.rb
fvzzy: I'm not sure I understand this comment. Isn't your suggestion what I'm already doing?

keitheous: Definitely. I was referring to the # good (normal indent) where we go

```
errors <<
something.method(
  key_1: '...',
  key_2: '...'
)
```

But yours look good too! :)

lib/zendesk_apps_support/package.rb
keitheous: 👍

config/locales/translations/zendesk_apps_support.yml
```diff
@@ -157,6 +165,14 @@ parts:
...
       title: "App builder job: <style> tags in template error"
       value: "<style> tag in %{template}. Use an app.css file instead."
+  - translation:
+      key: "txt.apps.admin.error.app_build.garden_cdn_reference"
+      title: "App builder job: Zendesk Garden CSS CDN reference in iframe template"
+      value: "Deprecated Zendesk Garden CSS referenced in %{file}. Update <link> to %{reference} to load from jsDelivr instead. See: https://jsdelivr.com/?query=zendeskgarden"
+  - translation:
+      key: "txt.apps.admin.error.app_build.legacy_sdk_url_reference"
+      title: "App builder job: legacy ZAF SDK url reference in iframe template"
+      value: "Deprecated Zendesk App Framework SDK referenced in %{file}. Update <script> for %{reference} to %{current_sdk_url}"
```
danielbreves: I usually prefer to point to the docs and be less specific here, in case the details change, it's easier to change the docs.

lib/zendesk_apps_support/package.rb
danielbreves: The name of this method seems to imply the app is all of those things at the same time.

lib/zendesk_apps_support/package.rb
```diff
@@ -31,19 +31,22 @@ def initialize(dir, is_cached = true)
...
       errors << Validations::Manifest.call(self)
+
       if has_valid_manifest?(errors)
         errors << Validations::Marketplace.call(self) if marketplace
         errors << Validations::Source.call(self)
         errors << Validations::Translations.call(self, skip_marketplace_translations: skip_marketplace_translations)
         errors << Validations::Requirements.call(self)
 
-        unless manifest.requirements_only? || manifest.marketing_only? || manifest.iframe_only?
+        unless requirements_marketing_iframe_only_app?(manifest)
```
danielbreves: `if v1_app`?

lib/zendesk_apps_support/validations/libraries.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+module ZendeskAppsSupport
+  module Validations
+    module Libraries
+      LEGACY_ASSETS = %r{https?://\S*assets\.zendesk\.com}
+      GARDEN_CDN_URL = /<link\s.*href="(#{LEGACY_ASSETS.source}\S*zendesk_garden\.css).*>/
+      LEGACY_SDK_URL = %r{<script\s.*src="(#{LEGACY_ASSETS.source}\S*\/((\d(\.\d+){1,2})|latest)\/zaf_sdk\.js).*>}
```
danielbreves: Feels slightly weird to include validations against legacy features. This code should eventually go away, right? Should we have a comment to explain why this is needed?

lib/zendesk_apps_support/package.rb
fvzzy: Not sure I understand this suggestion. If it's a `requirements_only` or `marketing_only` app it can still be "V2" right?

Agree that `requirements_marketing_iframe_only_app` isn't the most catchy method name, but I prefer that it's explicit

config/locales/translations/zendesk_apps_support.yml
fvzzy: Mm I'm not sure I agree. It feels like a nicer developer experience to me that if we're rejecting an app because of an incorrect reference that we then make it as easy as possible to correct that error (by including the link to the file here). 

It's a small PR to update docs, but equally it's only a single line here: https://github.com/zendesk/zendesk_apps_support/pull/217/files#diff-0770dbea04431e199f0b2b1f37d34610R37, and realistically this isn't likely to change often

lib/zendesk_apps_support/validations/libraries.rb
fvzzy: Yep this should absolutely be temporary, _but_ both old assets are going to hang around for a while and it should help to deprecate the legacy assets. I'll add a comment to explain this 👍

config/locales/translations/zendesk_apps_support.yml
```diff
@@ -131,12 +131,20 @@ parts:
...
+      title: "App builder job: banner image invalid format error"
       value: "Banner image must be a PNG file."
   - translation:
       key: "txt.apps.admin.error.app_build.banner.invalid_size"
-      title: "App builder job: Banner image invalid size error"
+      title: "App builder job: banner image invalid size error"
       value: "Invalid banner image dimensions. Must be %{required_banner_width}x%{required_banner_height}px."
+  - translation:
+      key: "txt.apps.admin.error.app_build.screenshots.excessive_files"
+      title: "App builder job: package contains too many screenshots"
```
hiroiguchi: Can you upload screenshot please?
Does the `%{file_count}` replaced with a number?
Also what will be inserted into `%{supported_count}`?

config/locales/translations/zendesk_apps_support.yml
```diff
@@ -157,6 +165,14 @@ parts:
       key: "txt.apps.admin.error.app_build.style_in_template"
       title: "App builder job: <style> tags in template error"
       value: "<style> tag in %{template}. Use an app.css file instead."
+  - translation:
+      key: "txt.apps.admin.error.app_build.garden_cdn_reference"
+      title: "App builder job: Zendesk Garden CSS CDN reference in iframe template"
+      value: "Deprecated Zendesk Garden CSS referenced in %{file}. Update <link> to %{reference} to load from jsDelivr instead. See: https://jsdelivr.com/?query=zendeskgarden"
```
hiroiguchi: Can you add screenshot please?

config/locales/translations/zendesk_apps_support.yml
hiroiguchi: Can you add screenshot please?

Mehul-Porwal:
@hiroiguchi Can you take another look at this PR? Thanks! 😄

config/locales/translations/zendesk_apps_support.yml
```diff
@@ -157,6 +167,16 @@ parts:
...
+      screenshot: "https://drive.google.com/file/d/1Waqh_DDHciuBznEnLJDq7eMnlUsPfHcX"
+  - translation:
+      key: "txt.apps.admin.error.app_build.legacy_sdk_url_reference"
+      title: "App builder job: legacy ZAF SDK url reference in iframe template"
```
hiroiguchi: Can you add following to the title please?

> %{file} is replaced with the name of a file. %{reference} and %{current_sdk_url} are replaced with the URL path.

lib/zendesk_apps_support/validations/screenshots.rb
```diff
@@ -0,0 +1,36 @@
...
+          end
+
+          screenshot_paths.each do |path|
+            screenshot = ImageSize.new(File.open(path))
+            next if screenshot.width == SCREENSHOT_WIDTH && screenshot.height == SCREENSHOT_HEIGHT
```
ocke: Wouldn't this limit all screenshots to be always exact 1024*768? Isn't that a bit limiting?

hiroiguchi:
@zendesk/localization

rdiaszen:
This PR seems to be stale. Is someone working on it?

rdiaszen:
Removing Localization team from this PR. It seems to be stale. 
Please if you need the strings to be reviewed ping us again.
Thank you.

fvzzy:
This is _so_ old. Closed as it's for a feature that likely doesn't even exist anymore.

PR #220, opened 2019-03-05
[AF-1509] Bump loofah and nokogiri to address security vulnerability

token-cjg:
### Context

A recent github security alerts report mentioned that loofah < 2.2.3 and nokogiri < 1.8.5 have known security vulnerabilities.  Therefore, per the recommendation of this report, bump loofah to 2.2.3 and nokogiri to 1.8.5.

### References

* https://zendesk.atlassian.net/browse/AF-1509
* https://github.com/flavorjones/loofah/compare/v2.2.1...v2.2.3
* https://github.com/sparklemotion/nokogiri/compare/v1.8.1...v1.8.5

### Risks

* (Medium) Breaks app validation in the apps_approval gem, which is used within Cactus (the app developer portal), thereby leading to problems with app submission
* (Low) Breaks app validations in the zendesk_apps_tools gem, thereby leading to issues with the zat command line tool

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -7,8 +7,8 @@ PATH
       image_size
       jshintrb (~> 0.3.0)
       json
-      loofah (~> 2.2.1)
-      nokogiri (~> 1.8.1)
+      loofah (~> 2.2.3)
+      nokogiri (~> 1.8.5)
       sass
       sassc (~> 1.11.2)
 
@@ -93,4 +93,4 @@ DEPENDENCIES
   zendesk_apps_support!
 
 BUNDLED WITH
-   1.16.2
+   1.16.3
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -47,8 +47,8 @@ GEM
       rake
     rake (12.3.0)
     rb-fsevent (0.10.3)
-    rb-inotify (0.10.0)
-      ffi (~> 1.0)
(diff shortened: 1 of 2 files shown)


token-cjg:
I'm getting a build error for Ruby 2.1:

> rb-inotify-0.10.0 requires ruby version >= 2.2, which is incompatible with the current version, ruby 2.1.10p492

Should we drop support for ruby 2.1.10p492, or should I try to downgrade rb-inotify in Gemfile.lock? @zendesk/vegemite

fvzzy:
Looks OK, but could we just update Nokogiri to the latest stable version (1.10, AFAIK) instead?

fvzzy:
Also could I suggest updating the risks in the description to fit a format that might be a bit more useful for other teams in case of any incidents, required rollback etc.? 

e.g. maybe something like:
_* Low: breaks app validations in Zendesk App Marketplace or Zendesk Apps Tools._

token-cjg:
hmmm

>nokogiri-1.10.1 requires ruby version >= 2.3.0, which is incompatible with the current version, ruby 2.2.6p396

Maybe we should just stick with nokogiri-1.8.5?

hoangtrvu:
Hey @token-cjg, so what's the minimum required Ruby version now after this change?

token-cjg:
The minimal required Ruby version will be 2.1.

PR #221, opened 2019-03-22
Remove signed setting uri validation

sarmadsangi:
✌️ @zendesk/vegemite

### Description
Remove validation that restricts `{{setting.yoursettingname}}` usage in signed urls.

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-1208

### Risks
* [Low] removes validation only

### DIFF
--- a/lib/zendesk_apps_support/validations/manifest.rb
+++ b/lib/zendesk_apps_support/validations/manifest.rb
@@ -273,6 +273,7 @@ module ZendeskAppsSupport
         def invalid_location_uri_error(package, location_options)
           path = location_options.url
           return nil if path == ZendeskAppsSupport::Manifest::LEGACY_URI_STUB
+          return nil if path.include?('{{setting.')
           validation_error = ValidationError.new(:invalid_location_uri, uri: path)
           uri = URI.parse(path)
           unless uri.absolute? ? valid_absolute_uri?(uri) : valid_relative_uri?(package, uri)
--- a/lib/zendesk_apps_support/validations/manifest.rb
+++ b/lib/zendesk_apps_support/validations/manifest.rb
@@ -273,9 +273,6 @@ module ZendeskAppsSupport
         def invalid_location_uri_error(package, location_options)
           path = location_options.url
           return nil if path == ZendeskAppsSupport::Manifest::LEGACY_URI_STUB
-          if path.include?('{{setting.')
-            return location_options.signed ? ValidationError.new(:signed_setting_uri, uri: path) : nil
-          end
           validation_error = ValidationError.new(:invalid_location_uri, uri: path)
           uri = URI.parse(path)
           unless uri.absolute? ? valid_absolute_uri?(uri) : valid_relative_uri?(package, uri)
--- a/spec/validations/manifest_spec.rb
+++ b/spec/validations/manifest_spec.rb
@@ -128,9 +128,11 @@ describe ZendeskAppsSupport::Validations::Manifest do
 
   it 'should not error when using {{setting.}}' do
     package = create_package('defaultLocale' => 'en')
-    allow(package.manifest.location_options.first).to receive(:url) { '{{setting.test}}' }
+    translation_files = double('AppFile', relative_path: 'translations/en.json')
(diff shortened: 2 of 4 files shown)


lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -273,9 +273,6 @@ def invalid_v1_location(package)
         def invalid_location_uri_error(package, location_options)
           path = location_options.url
           return nil if path == ZendeskAppsSupport::Manifest::LEGACY_URI_STUB
-          if path.include?('{{setting.')
-            return location_options.signed ? ValidationError.new(:signed_setting_uri, uri: path) : nil
```
hoangtrvu: can remove [this string](https://github.com/zendesk/zendesk_apps_support/blob/master/config/locales/en.yml#L83) too?

hoangtrvu:
cc @zendesk/wombat for review

lib/zendesk_apps_support/validations/manifest.rb
hoangtrvu: And [this string](https://github.com/zendesk/zendesk_apps_support/blob/917e290f5d7a4727122f1aa0e095658bcc77c8f9/config/locales/translations/zendesk_apps_support.yml#L193) too

sarmadsangi: Removed it locally but forgot to push this one out. Done now.

Btw, when do we remove this translation key? It won't be used, i guess i can remove it along with this change-set?
https://github.com/zendesk/zendesk_apps_support/blob/917e290f5d7a4727122f1aa0e095658bcc77c8f9/config/locales/en.yml#L83

hoangtrvu: Hang on, we shouldn't remove the whole block `if path.include?('{{setting.')` but should simply `return nil` instead right? Otherwise, it would continue to the check below where it tries to parse the path and check for absolute/relative url, which is not expected.

sarmadsangi: Good catch! Discussed in 🍖, it may break the app if we don't return `nil` here. I have added this condition back now with `return nil`.

sarmadsangi: Actually, would make sense to add one test for this scenario.

spec/validations/manifest_spec.rb
```diff
@@ -128,17 +128,11 @@ def create_package(parameter_hash)
...
-
-    expect(package).not_to have_error(/the url (.*) cannot use a setting because it is a signed url/)
-  end
-
-  it 'should error when using {{setting.}} with a signed url' do
-    package = create_package('defaultLocale' => 'en')
-    allow(package.manifest.location_options.first).to receive(:url) { '{{setting.test}}' }
-    allow(package.manifest.location_options.first).to receive(:signed) { true }
+    translation_files = double('AppFile', relative_path: 'translations/en.json')
+    allow(@package).to receive_messages(translation_files: [translation_files])
```
sarmadsangi: Question: is there a better way to bypass the missing translation file validation for this test?

spec/validations/manifest_spec.rb
```diff
@@ -128,17 +128,11 @@ def create_package(parameter_hash)
...
     package = create_package('defaultLocale' => 'en')
-    allow(package.manifest.location_options.first).to receive(:url) { '{{setting.test}}' }
-
-    expect(package).not_to have_error(/the url (.*) cannot use a setting because it is a signed url/)
-  end
-
-  it 'should error when using {{setting.}} with a signed url' do
-    package = create_package('defaultLocale' => 'en')
-    allow(package.manifest.location_options.first).to receive(:url) { '{{setting.test}}' }
-    allow(package.manifest.location_options.first).to receive(:signed) { true }
```
hoangtrvu: are you adding this test back to specifically test this behaviour for signed URL?

hoangtrvu: Actually ignore this, should not be necessary any more.

PR #223, opened 2019-04-09
Version lock dependencies

sarmadsangi:
✌️ @zendesk/vegemite 

### Summary

Version locking ZAS dependencies to avoid situations like this
https://travis-ci.org/zendesk/zendesk_apps_tools/jobs/517613519

- i18n version number was taken from ZAT
- Faker version number was taken from ZAM, also had to update this to make it work nicely with i18n version 1.5.1
- All others were taken from ZAS Gem.lock

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,14 +2,14 @@ PATH
   remote: .
   specs:
     zendesk_apps_support (4.15.0)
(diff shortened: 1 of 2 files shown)


keitheous:
This is now resolved. Closing

PR #224, opened 2019-05-02
[APPS-4597] Add plan support to installation

fbvilela:
@zendesk/wombat 

### Description

adds support to plan. To be used in ZAM in order to expose plan information to the framework

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -47,8 +47,8 @@ GEM
       rake
     rake (12.3.0)
     rb-fsevent (0.10.3)
-    rb-inotify (0.10.0)
-      ffi (~> 1.0)
+    rb-inotify (0.9.10)
+      ffi (>= 0.5.0, < 2)
     rspec (3.4.0)
       rspec-core (~> 3.4.0)
       rspec-expectations (~> 3.4.0)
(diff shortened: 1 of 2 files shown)


fbvilela:
https://docs.travis-ci.com/user/languages/ruby/#dependency-management 
looks like we have to drop the support for older versions of Ruby

I don't know how the latest pull request was green since it would have been after January 3rd 🤷‍♂ 

Can I just drop those ruby version in .travis.yml? 
@zendesk/vegemite @zendesk/wombat  ?

sarmadsangi:
@fbvilela we do support those versions in ZAT, so ideally, we would like to have these tests run against them to make sure we don't push out a version of ZAS that is incompatible with ZAT.

We are supposed to undertake some instrumentation work to figure out if those versions are still are being used. I guess after that we can drop support for older versions.

The issue that I faced was different, i ran into an issue with the dependency of ZAS in ZAT, because some of the dependencies in ZAS were not version locked.

fbvilela:
OK... well... I don't see how the two previous PRs could have been green. they would have faced the same problem. I'm keen to merge this as is... the code change is very small... is everyone ok with that?

PR #226, opened 2019-05-16
[MPORT-445] Warn when secrets are detected inside app text files

fvzzy:
🐕 

/cc @zendesk/dingo

### Description
Warn developers when the text files submitted in their app packages appear to contain secrets. Uses regexes from [the security checks PoC branch](https://github.com/zendesk/apps_approval/compare/fvilela/kari_security_work) to detect application specific secrets.

Currently these issues are caught as _warnings only_ and don't invalidate the app submission. 

#### Example ZAT output
<a href="https://cl.ly/37cea0ef143c" target="_blank"></a>

### References
* JIRA: https://zendesk.atlassian.net/browse/MPORT-445
* Translations: https://github.com/zendesk/zendesk_apps_support/pull/227

### Risks
* Low - causes alarm/confusion with developers previously unaware of exposed secrets in their code.

### DIFF
--- a/lib/zendesk_apps_support/package.rb
+++ b/lib/zendesk_apps_support/package.rb
@@ -3,6 +3,7 @@
 require 'pathname'
 require 'erubis'
 require 'json'
+require 'filemagic'
 
 module ZendeskAppsSupport
   class Package
@@ -33,6 +34,7 @@ module ZendeskAppsSupport
       errors << Validations::Manifest.call(self)
       if has_valid_manifest?(errors)
         errors << Validations::Marketplace.call(self) if marketplace
+        errors << Validations::Secrets.call(self)
         errors << Validations::Source.call(self)
         errors << Validations::Translations.call(self, skip_marketplace_translations: skip_marketplace_translations)
         errors << Validations::Requirements.call(self)
@@ -82,6 +84,10 @@ module ZendeskAppsSupport
         files << AppFile.new(self, relative_file_name)
       end
       files
+    end
+
+    def text_files
+      files.select { |f| text_file?(f) }
     end
 
     def js_files
@@ -364,6 +370,13 @@ module ZendeskAppsSupport
     def read_json(path, parser_opts = {})
       file = read_file(path)
       JSON.parse(read_file(path), parser_opts) unless file.nil?
+    end
+
+    def text_file?(file)
+      fm = FileMagic.new(FileMagic::MAGIC_MIME)
+      fm.file(file.absolute_path) =~ /^text\/|json/
+    ensure
+      fm.close
     end
   end
 end
--- a/lib/zendesk_apps_support/package.rb
+++ b/lib/zendesk_apps_support/package.rb
@@ -31,9 +31,9 @@ module ZendeskAppsSupport
     def validate(marketplace: true, skip_marketplace_translations: false)
       errors = []
       errors << Validations::Manifest.call(self)
+
       if has_valid_manifest?(errors)
         errors << Validations::Marketplace.call(self) if marketplace
-        errors << Validations::Secrets.call(self)
         errors << Validations::Source.call(self)
         errors << Validations::Translations.call(self, skip_marketplace_translations: skip_marketplace_translations)
         errors << Validations::Requirements.call(self)
@@ -46,6 +46,9 @@ module ZendeskAppsSupport
 
       errors << Validations::Banner.call(self) if has_banner?
       errors << Validations::Svg.call(self) if has_svgs?
+
+      # warning only validators
+      Validations::Secrets.call(self)
 
       errors.flatten.compact
     end
--- a/lib/zendesk_apps_support/package.rb
+++ b/lib/zendesk_apps_support/package.rb
@@ -86,7 +86,7 @@ module ZendeskAppsSupport
     end
 
     def text_files
-      @text_files ||= files.select { |f| f =~ %r{.*((ht|x)ml?|js(on)?)$} } # match .html, .xml, .js and .json
+      @text_files ||= files.select { |f| f =~ %r{.*(html?|xml|js|json?)$} }
     end
 
     def js_files
--- a/lib/zendesk_apps_support/package.rb
+++ b/lib/zendesk_apps_support/package.rb
@@ -3,7 +3,6 @@
 require 'pathname'
 require 'erubis'
 require 'json'
-require 'filemagic'
 
 module ZendeskAppsSupport
   class Package
@@ -87,7 +86,7 @@ module ZendeskAppsSupport
     end
 
     def text_files
-      files.select { |f| text_file?(f) }
+      @text_files ||= files.select { |f| f =~ %r{.*((ht|x)ml?|js(on)?)$} } # match .html, .xml, .js and .json
     end
 
     def js_files
@@ -370,13 +369,6 @@ module ZendeskAppsSupport
     def read_json(path, parser_opts = {})
       file = read_file(path)
       JSON.parse(read_file(path), parser_opts) unless file.nil?
-    end
-
-    def text_file?(file)
-      fm = FileMagic.new(FileMagic::MAGIC_MIME)
-      fm.file(file.absolute_path) =~ /^text\/|json/
-    ensure
-      fm.close
     end
   end
 end
--- a/lib/zendesk_apps_support/package.rb
+++ b/lib/zendesk_apps_support/package.rb
@@ -90,7 +90,7 @@ module ZendeskAppsSupport
     end
 
     def js_files
-      @js_files ||= files.select { |f| f.to_s == 'app.js' || (f.to_s.start_with?('lib/') && f.to_s.end_with?('.js')) }
+      @js_files ||= files.select { |f| f =~ %r{^(app|lib\/.*)\.js$} }
     end
 
     def lib_files
--- a/lib/zendesk_apps_support/validations/secrets.rb
+++ b/lib/zendesk_apps_support/validations/secrets.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+require 'net/http'
+require 'json'
+
+module ZendeskAppsSupport
+  module Validations
+    module Secrets
+      SECRET_KEYWORDS = %w[
+        pass password secret secretToken secret_token auth_key
+        authKey auth_pass authPass auth_user AuthUser username api_key
+      ].freeze
+
+      class << self
+        def call(package)
+          compromised_files = []
+
+          # Rely on the regexes used by truffleHog (https://github.com/dxa4481/truffleHog) to detect application secrets
+          trufflehog_regexes = JSON.parse(Net::HTTP.get(URI.parse('https://raw.githubusercontent.com/dxa4481/truffleHogRegexes/master/truffleHogRegexes/regexes.json'))).freeze
+
+          package.text_files.each do |file|
+            contents = file.read
+
+            trufflehog_regexes.each do |secret_type, regex_str|
+              next unless contents =~ Regexp.new(regex_str)
+              package.warnings << I18n.t('txt.apps.admin.warning.app_build.application_secret',
+                                         file: file,
+                                         secret_type: secret_type)
+            end
+
+            compromised_files << file.relative_path if contents =~ Regexp.union(SECRET_KEYWORDS)
+          end
+
+          if compromised_files
+            package.warnings << I18n.t('txt.apps.admin.warning.app_build.generic_secrets',
+                                       files: compromised_files.join(', '),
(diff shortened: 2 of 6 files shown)


lib/zendesk_apps_support/package.rb
```diff
@@ -365,5 +371,12 @@ def read_json(path, parser_opts = {})
       file = read_file(path)
       JSON.parse(read_file(path), parser_opts) unless file.nil?
     end
+
+    def text_file?(file)
+      fm = FileMagic.new(FileMagic::MAGIC_MIME)
+      fm.file(file.absolute_path) =~ /^text\/|json/
```
fvzzy: I'm not really sure about this - should I instead filter out `charset=binary` in this string?

lib/zendesk_apps_support/validations/secrets.rb
```diff
@@ -0,0 +1,44 @@
...
+        pass password secret secretToken secret_token auth_key
+        authKey auth_pass authPass auth_user AuthUser username api_key
+      ].freeze
+
+      class << self
+        def call(package)
+          compromised_files = []
+
+          # Rely on the regexes used by truffleHog (https://github.com/dxa4481/truffleHog) to detect application secrets
+          trufflehog_regexes = JSON.parse(Net::HTTP.get(URI.parse('https://raw.githubusercontent.com/dxa4481/truffleHogRegexes/master/truffleHogRegexes/regexes.json'))).freeze
```
fvzzy: Seems silly to make a network request for this at runtime, though I wasn't sure of the correct way to include these when compiling the gem. Open to suggestions here.

seangoedecke: We could patch `bump/tasks` to download a new version of the file when a new version of the gem is bumped. I'm in favour of hardcoding the file anyway so that if that download fails we at least have something.

lib/zendesk_apps_support/validations/secrets.rb
```diff
@@ -0,0 +1,44 @@
...
+
+          package.text_files.each do |file|
+            contents = file.read
+
+            trufflehog_regexes.each do |secret_type, regex_str|
```
seangoedecke: Can we `Regexp.union` these regexes? We'd lose the ability to report on what type of secret was found, but it'd probably be obvious to the developer, right?

fvzzy: Doing that seems like a very minor increase in code length + speed to validate vs a potentially very useful pointer to the secret for the app developer. I'd prefer to keep the type information in here.

seangoedecke: Sure, I agree that this does provide more information to the app developer. Happy to keep it in here.

lib/zendesk_apps_support/package.rb
seangoedecke: Do we need to introduce a dependency on filemagic for this? What if instead we just matched on `.js, .html, .json` file extensions? Would we miss out any files that we care about?

Introducing a dependency on an external C library risks setup problems, especially since this gem is being included in a cross-platform CLI tool (ZAT)

lib/zendesk_apps_support/package.rb
```diff
@@ -365,5 +371,12 @@ def read_json(path, parser_opts = {})
       file = read_file(path)
       JSON.parse(read_file(path), parser_opts) unless file.nil?
     end
+
+    def text_file?(file)
```
seangoedecke: nit: can we push this check down into `validations/secrets.rb`, since it's not being used anywhere else?

lib/zendesk_apps_support/package.rb
fvzzy: Yeah, I'm not sure. I originally wanted to scan all files that aren't binaries, but perhaps those file extensions are enough 🤷‍♀

FileMagic is probably a little heavy-handed in any case

lib/zendesk_apps_support/package.rb
fvzzy: I wanted `text_files` to live alongside its other file filter method friends in this class (and so also defined `text_file?` in the same file), but given https://github.com/zendesk/zendesk_apps_support/pull/226/files/d28b0b1cd7ad21a48bc8026807c3edc762c5a46a#r284574646 I'll probably just remove this altogether

config/locales/translations/zendesk_apps_support.yml
```diff
@@ -219,6 +219,18 @@ parts:
...
+      key: "txt.apps.admin.warning.app_build.generic_secrets.one"
+      title: "App builder job: warning for generic secrets found in app text files"
+      value: "Possible secrets found in %{files}. Consider reviewing the contents of this file before submitting your app."
+  - translation:
+      key: "txt.apps.admin.warning.app_build.generic_secrets.other"
+      title: "App builder job: warning for generic secrets found in app text files"
+      value: "Possible secrets found in %{files}. Consider reviewing the contents of these files before submitting your app."
+  - translation:
+      key: "txt.apps.admin.warning.app_build.application_secret"
+      title: "App buider job: warning for secrets found in app text files"
```
seangoedecke: nit: builder

lib/zendesk_apps_support/package.rb
```diff
@@ -84,8 +85,12 @@ def files
       files
     end
 
+    def text_files
+      @text_files ||= files.select { |f| f =~ %r{.*((ht|x)ml?|js(on)?)$} } # match .html, .xml, .js and .json
```
seangoedecke: This is very cute but I wonder if it'd be more readable as `(html|xml|js|json)$`?

lib/zendesk_apps_support/package.rb
```diff
@@ -84,8 +85,12 @@ def files
...
+    end
+
     def js_files
-      @js_files ||= files.select { |f| f.to_s == 'app.js' || (f.to_s.start_with?('lib/') && f.to_s.end_with?('.js')) }
```
seangoedecke: I have a preference for the code as it was (i.e. not using a regex until it's the only option), but happy to go with this if others think it's nicer

lib/zendesk_apps_support/validations/secrets.rb
```diff
@@ -0,0 +1,78 @@
...
+        'Password' => /([pP][aA][sS][sS][wW][oO][rR][dD].*.[0-9a-zA-Z+_.-]{4,156})/,
+        'Token' => /([tT][oO][kK][eE][nN]( [:\"=-]|[:\"=-]).*.[0-9a-zA-Z+_.-]{4,156})/
+        # rubocop:enable Metrics/LineLength
+      }.freeze
+
+      class << self
+        def call(package)
+          compromised_files = []
+
+          package.text_files.each do |file|
```
seangoedecke: slight preference for `compromised_files = package.text_files.map do |file|... end.compact` instead of this pattern

spec/validations/secrets_spec.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+json_base = '{"name":"App","author":{},"location":{},"version":"1.0"}'
+html_base = '<html><body><script src="zaf_sdk.js"></script></body></html>'
+js_base = 'const client = ZAFClient.init();'
+
+describe ZendeskAppsSupport::Validations::Svg do
```
seangoedecke: nit: should be `Secrets` instead of `Svg`

spec/validations/secrets_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  let(:mock_html) { html_base }
+  let(:mock_js) { js_base }
+  let(:subject) { ZendeskAppsSupport::Validations::Secrets }
+  let(:manifest_json) { double('AppFile', relative_path: 'manifest.json', read: mock_json) }
+  let(:iframe_html) { double('AppFile', relative_path: 'assets/iframe.html', read: mock_html) }
+  let(:app_js) { double('AppFile', relative_path: 'assets/app.js', read: mock_js) }
+  let(:files) { [manifest_json, iframe_html, app_js] }
+  let(:package) { double('Package', text_files: files, warnings: []) }
+
+  context 'a single text file containing generic secret keyword(s)' do
```
seangoedecke: Really nice tests 👌

lib/zendesk_apps_support/package.rb
fvzzy: Fine.

lib/zendesk_apps_support/package.rb
fvzzy: Yeah, fair. Just did this for consistency with the matchers surrounding it but not bothered either way

lib/zendesk_apps_support/package.rb
fvzzy: I mean my regex would've matched on `.xm` files as well for [Triton FastTracker 2 extended modules](https://en.wikipedia.org/wiki/XM_(file_format) containing exposed secrets, but we'll just ignore that feature then

lib/zendesk_apps_support/validations/secrets.rb
fvzzy: Haven't done this before - this is nice 👍

lib/zendesk_apps_support/validations/secrets.rb
```diff
@@ -0,0 +1,76 @@
...
+
+            file.relative_path if contents =~ Regexp.union(SECRET_KEYWORDS)
+          end.compact
+
+          if compromised_files.any?
+            package.warnings << I18n.t('txt.apps.admin.warning.app_build.generic_secrets',
+                                       files: compromised_files.join(', '),
+                                       count: compromised_files.count)
+          end
+          []
```
keitheous: Is there a reason why we return an empty array here? Should we be returning package or package.warning?

keitheous: In the test, we assign

```
errors = subject.call(package)
```

Since the method returns an empty array, does it mean that expect(errors).to be_empty will always be the case?

fvzzy: Ah yeah so admittedly this isn't super clear - in https://github.com/zendesk/zendesk_apps_support/blob/master/lib/zendesk_apps_support/package.rb#L31, we collate errors from each of these "validator" classes. Each validator returns an array of errors, but since this one only displays warnings (no 'blocking' errors) I just set the return value to an empty array here.

I could instead assign `errors = []` at the start of this file (e.g. https://github.com/zendesk/zendesk_apps_support/blob/master/lib/zendesk_apps_support/validations/svg.rb#L56) and return `errors` at the end if you think that's better?

seangoedecke: If we're never appending to errors I think it could be misleading to have that variable. What about adding a comment `[] // no errors` or something?

fvzzy: We actually do something similar in the `banner` validator (https://github.com/zendesk/zendesk_apps_support/blob/master/lib/zendesk_apps_support/validations/banner.rb#L30), but it is kind of confusing.

keitheous: As discussed in person, we decided to perhaps call the method on its own. eg

`errors << validate_no_secrets()` vs `validate_no_secrets()`

as `package.warnings` needs not append errors collection

config/locales/translations/zendesk_apps_support.yml
```diff
@@ -219,6 +219,18 @@ parts:
       key: "txt.apps.admin.warning.app_build.bitmap_in_svg"
       title: "App builder job: warning that svg contains an embedded bitmap image and cannot be used"
       value: "%{svg} contains an embedded bitmap and cannot be used as an app icon. It has been replaced with a default placeholder icon."
+  - translation:
```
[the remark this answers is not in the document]
fvzzy: Argh yeah good catch, I always forget this. Is there an example of a translations only PR about?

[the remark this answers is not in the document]
fvzzy: https://github.com/zendesk/zendesk_apps_support/pull/227

lib/zendesk_apps_support/validations/secrets.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+module ZendeskAppsSupport
+  module Validations
+    module Secrets
+      SECRET_KEYWORDS = %w[
+        pass password secret secretToken secret_token auth_key
```
mtsolakiszen: Just something worth noting, the matcher for these keywords just checks if they're present in the file.
While testing this gave a false positive as `passed` was present in a file which matched against `pass`. Just a small thing worth noting which could annoy devs but definitely not a showstopper imo.

PR #227, opened 2019-05-23
[MPORT-445] Add strings for secret warnings

fvzzy:
🐕 

/cc @zendesk/dingo @zendesk/localization 

### Description
Strings for https://github.com/zendesk/zendesk_apps_support/pull/226.

#### Example ZAT output
<a href="https://cl.ly/37cea0ef143c" target="_blank"></a>

### References
* JIRA: https://zendesk.atlassian.net/browse/MPORT-445

### DIFF
--- a/config/locales/translations/zendesk_apps_support.yml
+++ b/config/locales/translations/zendesk_apps_support.yml
@@ -220,6 +220,30 @@ parts:
       title: "App builder job: warning that svg contains an embedded bitmap image and cannot be used"
       value: "%{svg} contains an embedded bitmap and cannot be used as an app icon. It has been replaced with a default placeholder icon."
   - translation:
+      key: "txt.apps.admin.warning.app_build.generic_secrets.one"
+      title: "App builder job: warning for generic secrets found in app text files"
+      value: "Possible secrets found in %{files}. Consider reviewing the contents of this file before submitting your app."
+  - translation:
+      key: "txt.apps.admin.warning.app_build.generic_secrets.other"
+      title: "App builder job: warning for generic secrets found in app text files"
+      value: "Possible secrets found in %{files}. Consider reviewing the contents of these files before submitting your app."
+  - translation:
+      key: "txt.apps.admin.warning.app_build.application_secret"
+      title: "App builder job: warning for secrets found in app text files"
+      value: "Possible %{secret_type} found in %{file}. Consider reviewing the contents of this file before submitting your app."
+  - translation:
+      key: "txt.apps.admin.warning.app_build.generic_secrets.one"
+      title: "App builder job: warning for generic secrets found in app text files"
+      value: "Possible secrets found in %{files}. Consider reviewing the contents of this file before submitting your app."
+  - translation:
+      key: "txt.apps.admin.warning.app_build.generic_secrets.other"
+      title: "App builder job: warning for generic secrets found in app text files"
+      value: "Possible secrets found in %{files}. Consider reviewing the contents of these files before submitting your app."
+  - translation:
+      key: "txt.apps.admin.warning.app_build.application_secret"
+      title: "App builder job: warning for secrets found in app text files"
+      value: "Possible %{secret_type} found in %{file}. Consider reviewing the contents of this file before submitting your app."
+  - translation:
       key: "txt.apps.admin.error.app_build.invalid_version"
       title: "App builder job: invalid framework version"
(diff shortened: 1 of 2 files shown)


config/locales/translations/zendesk_apps_support.yml
```diff
@@ -219,6 +219,30 @@ parts:
       key: "txt.apps.admin.warning.app_build.bitmap_in_svg"
       title: "App builder job: warning that svg contains an embedded bitmap image and cannot be used"
       value: "%{svg} contains an embedded bitmap and cannot be used as an app icon. It has been replaced with a default placeholder icon."
+  - translation:
+      key: "txt.apps.admin.warning.app_build.generic_secrets.one"
+      title: "App builder job: warning for generic secrets found in app text files"
+      value: "Possible secrets found in %{files}. Consider reviewing the contents of this file before submitting your app."
```
hiroiguchi: According to the screenshots, the `%{files}` is replaced with multiple files such as `manifest.json, assets/background.html`. The `,` needs to be localized. Is this extracted?

e.g: `manifest.json, assets/background.html` will be `manifest.json、assets/background.html` in Japanese.
I would like linguist to localze `,` and let them to choose inserting a space or not. Above Japanese case, we don't insert a space but only `、`.

config/locales/translations/zendesk_apps_support.yml
```diff
@@ -219,6 +219,30 @@ parts:
...
+  - translation:
+      key: "txt.apps.admin.warning.app_build.generic_secrets.other"
+      title: "App builder job: warning for generic secrets found in app text files"
+      value: "Possible secrets found in %{files}. Consider reviewing the contents of these files before submitting your app."
```
hiroiguchi: Can you upload screenshots to Google Drive so that linguists will be able to see please?

https://zendesk.atlassian.net/wiki/spaces/globalization/pages/401277972/Upload+screenshots+to+Google+Drive

config/locales/translations/zendesk_apps_support.yml
fvzzy: Hey @hiroiguchi, thanks for pointing this out.

I can extract out the comma into a seperate string, but as this is a pretty common pattern (i.e. joining a list of things with `', '`) is there a general purpose string I could pull in instead? It seems like this could be reused and perhaps like this shouldn't be specific to this project namespace `txt.apps.admin.warning.app_build.x`

hiroiguchi: I found some comma strings like these 

https://github.com/zendesk/zendesk/blob/10282076a73c252a1eeefab90576b0e9d130ba93/config/locales/translations/agent_classic.yml#L1074-L1076
https://github.com/zendesk/zendesk_console/blob/master/lotus_react/translations/default.yml#L791-L793

However I would recommend you to create the one for app because the comma can be different with context. 
For example, I would use `、` in Japanese for this case but if you are listing numbers, I would use `,`.  Using key from other repos may cause issues.

config/locales/translations/zendesk_apps_support.yml
```diff
@@ -219,6 +223,21 @@ parts:
       key: "txt.apps.admin.warning.app_build.bitmap_in_svg"
       title: "App builder job: warning that svg contains an embedded bitmap image and cannot be used"
       value: "%{svg} contains an embedded bitmap and cannot be used as an app icon. It has been replaced with a default placeholder icon."
+  - translation:
+      key: "txt.apps.admin.warning.app_build.generic_secrets.one"
+      title: "App builder job: warning for generic secrets found in app text files"
```
hiroiguchi: Can you add following to the title please?

> %{files} will be replaced with one file name.

config/locales/translations/zendesk_apps_support.yml
```diff
@@ -219,6 +223,21 @@ parts:
...
+      value: "Possible secrets found in %{files}. Consider reviewing the contents of this file before submitting your app."
+      screenshot: "https://drive.google.com/file/d/179IMwzJvXD1m5u-0K499Ul3-qRBXAxnT"
+  - translation:
+      key: "txt.apps.admin.warning.app_build.generic_secrets.other"
+      title: "App builder job: warning for generic secrets found in app text files"
```
hiroiguchi: Can you add following to the title please?

> %{files} will be replaced with multiple file name. If you language has more than one plural, use Possible secrets found in files: %{files}. Consider reviewing the contents of these files before submitting your app.

hiroiguchi:
👌 for strings!

PR #228, opened 2019-06-04
[MPORT-442] Validate MIME types for Security Checks

keitheous:
@zendesk/dingo  @zendesk/wattle-dev 

# Description
As discussed in our security checks sessions, we thought that it would be more reasonable to put this validation check earlier on in the app creation process - in ZAS, rather than in Apps Approval. And as a result, this security check can then be used as a **blocking error** in

1) ZAT - to package and validate - checkout `zat validate` / `zat package` in command.rb

2) Apps Approval (consumed by ZAM) when Uploading Apps - checkout `AppsApproval::AppLoader.new({dir: dir, zip: zip}).load_app`

3) ZAM - when Building Apps - TBC: Will find out how warnings/errors go about end to end

# TODO
- [x] Create pre-requisite branch to  I18n validation error string
- [x] https://github.com/zendesk/zendesk_apps_support/pull/229 is approved and merged?
- [x] Wait for Rosetta Sweep
- [x] Include test
- [x] Update test to use Localization
- [x] Figure out a better warning string than "Excuse me! File ..."
- [ ] Seriously, checkout how ZAM uses this for to Build Apps

# Approach
1) Security team left us a list of block listed mime types, they are 
    ```
     UNSUPPORTED_MIME_TYPES = %w[
        vnd.rar rar zip gzip pdf doc docx avi bin bz bz2 csh sh jar mp3 mpeg odt pptx ppt xls xlsx 7z
     ].freeze
    ```

    Reverting back to Mime Magic Gem to identify the real content of a file. Mime Magic opens up the 
    file and read the contents.

2) Mime Magic is used to identify the real content mime type and the extension name. If either is identified as a block listed mime type, we spit (pui pui*) an error to inform the users that there are files with unsupported mime types in their application.

3) Notes: The reason why I decided to check both real content mime type and extension name is because doc/docx are identified as msword. However the list above has doc/docx as unsupported mime type.
    Screen Shot 2019-06-12 at 10 07 58 am
    Screen Shot 2019-06-12 at 10 12 12 am

# References
- [JIRA](https://zendesk.atlassian.net/browse/MPORT-442)
- [Kari's Security Work](https://github.com/zendesk/apps_approval/compare/master...fvilela/kari_security_work)

# Risks
- Low: validation task that isn't called doesn't execute?


lib/zendesk_apps_support/validations/mime.rb
```diff
@@ -0,0 +1,42 @@
...
+          app_file_mime_type = identify_mime_type(app_file)
+
+          file_extension = app_file_mime_type.extension
+          content_subtype = app_file_mime_type.content_type.split('/').last
+
+          ([file_extension, content_subtype] & UNSUPPORTED_MIME_TYPES).any?
+        end
+
+        def identify_mime_type(app_file)
+          require 'mini_mime'
```
seangoedecke: nit: I prefer to have this at the top of the file, so you don't have to read through the code to see dependencies

lib/zendesk_apps_support/validations/mime.rb
```diff
@@ -0,0 +1,42 @@
...
+  module Validations
+    module Mime
+      UNSUPPORTED_MIME_TYPES = %w[
+        vnd.rar rar zip gzip pdf doc docx avi bin bz bz2 csh sh jar mp3 mpeg odt pptx ppt xls xlsx 7z
+      ].freeze
+
+      class << self
+        def call(package)
+          unsupported_files = package.files.find_all do |app_file|
+            identify_mime_type(app_file) && block_listed?(app_file)
```
seangoedecke: Why do we need `identify_mime_type(app_file) &&` here?

lib/zendesk_apps_support/validations/mime.rb
seangoedecke: In fact, could this method just be collapsed into `block_listed?`, since it's just an alias for a `MiniMime` method?

lib/zendesk_apps_support/validations/mime.rb
```diff
@@ -0,0 +1,42 @@
...
+          end.map(&:relative_path)
+
+          [mime_type_warning(unsupported_files)] if unsupported_files.any?
+        end
+
+        private
+
+        def block_listed?(app_file)
+          app_file_mime_type = identify_mime_type(app_file)
```
seangoedecke: nit: could just call this variable `mime_type` and save some reading

spec/validations/mime_spec.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe ZendeskAppsSupport::Validations::Mime do
```
seangoedecke: Nice tests!

lib/zendesk_apps_support/validations/mime.rb
keitheous: Because when we were using MimeMagic, we identify that gems might not be able to identify all mime types. 

I was thinking, if they are not identify-able as one of the UNSUPPORTED_MIME_TYPES, it shouldn't matter. Also works as a precaution. 

Curious to know what do you think?

lib/zendesk_apps_support/validations/mime.rb
keitheous: YES! I like!

lib/zendesk_apps_support/validations/mime.rb
keitheous: YES we can! It doesn't serve much purpose being extracted out as a private method anyway. Love the suggestion!!!

spec/validations/mime_spec.rb
keitheous: Thank you, Sean!

lib/zendesk_apps_support/validations/mime.rb
keitheous: As discussed in meat space,

we have decided to do an
`early return if mime type is mime_type.nil?` or `early return unless mime_type`

keitheous: Update:

Reverting back to MimeMagic because it opens files to evaluate content.
```
mime_type = MimeMagic.by_magic(app_file.read)

content_subtype = mime_type.subtype if mime_type
extension_name = app_file.extension.delete('.')

([content_subtype, extension_name] & UNSUPPORTED_MIME_TYPES).any?
```
With the lines above, we do not need to worry if content subtype can't be identified, content_subtype can be nil. Also, we check for `app_file.extension` as well.

spec/validations/mime_spec.rb
```diff
@@ -0,0 +1,51 @@
...
```
fvzzy: Nice work. Perhaps add a test (or two) to validate against a couple of block-listed files masked as safe ones?

keitheous: Thank you for the suggestion. I have added that.

I included a zip file that i renamed as something.html (in spec/fixtures)

MimeMagic still identifies it as a zip
Screen Shot 2019-06-13 at 11 41 49 am

lib/zendesk_apps_support/validations/mime.rb
```diff
@@ -0,0 +1,43 @@
...
+          mime_type = MimeMagic.by_magic(app_file.read)
+
+          content_subtype = mime_type.subtype if mime_type
+          extension_name = app_file.extension.delete('.')
+
+          ([content_subtype, extension_name] & UNSUPPORTED_MIME_TYPES).any?
+        end
+
+        def mime_type_warning(file_names)
+          # TODO: replace string with block below - pluralism is handled - reminder to update tests
```
mtsolakiszen: Reminder this needs to be done once translations PR is merged

mtsolakiszen:
Tested locally. Works fine and error messages shown correctly in ZAM

PR #230, opened 2019-06-21
[MPORT-443] Add validations for insecure and block-listed requests

fvzzy:
🐕 

/cc @zendesk/dingo

### Description
1. Scan JS and HTML files for common Javascript request function call patterns (i.e. jQuery, XMLHttpRequest, fetch, and Zendesk App Framework `.request` calls)
1. **Error** if calls are being made to:
	1. RFC1918 **private** network addresses (10.0.0.0/8, 172.16.0.0/12, or 192.168.0.0/16 range)
	1. RFC1122 **loopback** network addresses (127.0.0.0/8 range)
	1. RFC3927 **link-local** network addresses (169.254.0.0/16 range)
1. **Warn** if requests are made via `http` instead of `https`.

Uses the [IPAddress gem](https://github.com/ipaddress-gem/ipaddress) to validate IPs and address types.

### References
* JIRA: https://zendesk.atlassian.net/browse/MPORT-443
* Translations: https://github.com/zendesk/zendesk_apps_support/pull/231

### Risks
* Medium - causes apps with legitimate requests to be rejected.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -5,6 +5,7 @@ PATH
       erubis
       i18n
       image_size
+      ipaddress_2 (~> 0.13.0)
       json
       loofah (~> 2.2.3)
       mimemagic (~> 0.3.3)
@@ -27,6 +28,7 @@ GEM
     ffi (1.9.25)
     i18n (0.7.0)
     image_size (2.0.1)
+    ipaddress_2 (0.13.0)
     json (2.2.0)
     loofah (2.2.3)
       crass (~> 1.0.2)
--- a/lib/zendesk_apps_support/package.rb
+++ b/lib/zendesk_apps_support/package.rb
@@ -37,6 +37,7 @@ module ZendeskAppsSupport
         errors << Validations::Source.call(self)
         errors << Validations::Translations.call(self, skip_marketplace_translations: skip_marketplace_translations)
         errors << Validations::Requirements.call(self)
+        errors << Validations::Requests.call(self)
 
         unless manifest.requirements_only? || manifest.marketing_only? || manifest.iframe_only?
           errors << Validations::Templates.call(self)
@@ -95,6 +96,10 @@ module ZendeskAppsSupport
 
     def js_files
       @js_files ||= files.select { |f| f =~ %r{^.*\.jsx?$} }
+    end
+
+    def html_files
+      @html_files ||= files.select { |f| f =~ %r{.*\.html?$} }
     end
 
     def lib_files
--- a/lib/zendesk_apps_support/validations/requests.rb
+++ b/lib/zendesk_apps_support/validations/requests.rb
@@ -1,0 +1,58 @@
+# frozen_string_literal: true
+
+require 'uri'
+require 'ipaddress_2'
+
+module ZendeskAppsSupport
+  module Validations
+    module Requests
+      class << self
+        HTTP_REQUEST_METHODS = %w[GET HEAD POST PUT DELETE CONNECT OPTIONS TRACE].freeze
+
+        REQUEST_CALLS = [
+          /\w+\.request\(['"](.*)['"]/i, # ZAF request
+          /(?:\$|jQuery)\.(?:ajax|get|post|getJSON)\(['"](.*)['"]/i, # jQuery request
+          /\w+\.open\(['"](?:#{Regexp.union(HTTP_REQUEST_METHODS).source})['"],\s?['"](.*)['"]/i, # XMLHttpRequest
+          /fetch\(['"](.*)['"]/i # fetch
+        ].freeze
+
+        def call(package)
+          errors = []
+          files = package.js_files + package.html_files
+
+          files.each do |file|
+            contents = file.read
+            REQUEST_CALLS.each do |request_pattern|
+              request = contents.match(request_pattern)
+              next unless request
+              uri = URI(request.captures[0])
+              if uri.scheme == 'http'
+                package.warnings << I18n.t('txt.apps.admin.warning.app_build.insecure_http_request',
+                                           uri: uri,
+                                           file: file.relative_path)
+              end
+
+              next unless IPAddress.valid? uri.host
+              ip = IPAddress.parse uri.host
+
+              blocked_ip_type = if ip.private?
+                                  I18n.t('txt.apps.admin.error.app_build.blocked_request_private')
+                                elsif ip.loopback?
+                                  I18n.t('txt.apps.admin.error.app_build.blocked_request_loopback')
+                                elsif ip.link_local?
+                                  I18n.t('txt.apps.admin.error.app_build.blocked_request_link_local')
+                                end
+
+              next unless blocked_ip_type
+              errors << I18n.t('txt.apps.admin.error.app_build.blocked_request',
+                               type: blocked_ip_type,
+                               uri: uri.host,
+                               file: file.relative_path)
+            end
+          end
+          errors
+        end
+      end
+    end
+  end
+end
--- a/spec/validations/requests_spec.rb
+++ b/spec/validations/requests_spec.rb
@@ -1,0 +1,76 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+def request_function(style, address)
+  case style
+  when 'ZAF'
+    "function request() { client.request('#{address}') }"
+  when 'jQuery'
+    "function request() { jQuery.get('#{address}') }"
+  when 'jQuery$'
+    "function request() { $.ajax('#{address}') }"
+  when 'XMLHttpRequest'
+    "function request() { xhr.open('get', '#{address}') }"
+  when 'fetch'
+    "function request() { fetch('#{address}') }"
+  end
+end
+
+request_function_styles = %w[ZAF jQuery jQuery$ XMLHttpRequest fetch]
+
+shared_examples 'an insecure request' do |file_path, function_style|
+  address = 'http://foo.com'
+  let(:markup) { request_function(function_style, address) }
+
+  it "and raise a warning inside #{function_style} style requests" do
+    errors = subject.call(package)
+    expect(package.warnings[0]).to include('insecure HTTP request', address, file_path)
+    expect(errors).to be_empty
+  end
+end
+
+blocked_ips = {
+  private: {
+    range: '10.0.0.0/8, 172.16.0.0/12, or 192.168.0.0/16',
+    example: '192.168.0.1'
+  },
+  loopback: {
+    range: '127.0.0.0/8',
+    example: '127.0.0.1'
+  },
+  link_local: {
+    range: '169.254.0.0/16',
+    example: '169.254.0.1'
+  }
+}
+
+shared_examples 'a blocked ip' do |file_path, function_style, ip_type, ip|
+  let(:markup) { request_function(function_style, "https://#{ip}") }
+
+  it "and throw a #{ip_type} ip error inside #{function_style} style request calls" do
+    errors = subject.call(package)
+    expect(package.warnings).to be_empty
+    expect(errors[0]).to include("request to a #{ip_type} ip", ip, file_path)
+  end
+end
+
+describe ZendeskAppsSupport::Validations::Requests do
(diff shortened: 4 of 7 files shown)


lib/zendesk_apps_support/validations/requests.rb
```diff
@@ -0,0 +1,53 @@
...
+          files.each do |file|
+            contents = file.read
+            REQUEST_CALLS.each do |request_pattern|
+              request = contents.match(request_pattern)
+              next unless request
+              uri = URI(request.captures[0])
+              package.warnings << "Possible insecure HTTP request to #{uri} in #{file.relative_path}. Consider instead using the HTTPS protocol." if uri.scheme == 'http'
+
+              next unless IPAddress.valid? uri.host
+              ip = IPAddress uri.host
```
seangoedecke: What's going on here, syntactically? Never seen calling a class like a function before.

spec/validations/requests_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+  end
+end
+
+describe ZendeskAppsSupport::Validations::Requests do
+  let(:app_js) { double('AppFile', relative_path: 'assets/app.js', read: markup) }
+  let(:subject) { ZendeskAppsSupport::Validations::Requests }
+  let(:package) { double('Package', js_files: [app_js], html_files: [], warnings: []) }
+
+  describe 'using the http protocol' do
```
seangoedecke: I like how a bunch of this stuff is abstracted into shared examples. I wonder if it can be taken further - something like:
```
request_function_styles.each do |style|
  describe "#{style} type requests" do
    describe 'to some address range' do
      it_behaves_like 'a blocked X request', style
...
```

spec/validations/requests_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+  },
+  { type: 'link-local',
+    example_address: 'https://169.254.0.1',
+    # TODO: update with translated string
+    error: 'Possible request to a link-local network address 169.254.0.1 in assets/app.js'
+  }
+]
+
+blocked_requests.each do |blocked_request|
+  shared_examples "a blocked #{blocked_request[:type]} request" do |function_style|
```
seangoedecke: This makes me a bit uncomfortable because of the way it forces the test to rely on dynamically-generated magic strings that have to be kept in sync (e.g. modifying a request style means updating the object above, the array of request styles, and the middle of the test description string). I don't have any good suggestions here and I agree that it's better than all the duplication that would be involved in avoiding shared examples entirely.

What if we just had the one `a blocked request` shared example and passed in the type/markup/error as params?

spec/validations/requests_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+    error: 'Possible request to private network address 192.168.0.1 in assets/app.js'
+  },
+  {
+    type: 'loopback',
+    example_address: 'https://127.0.0.1',
+    # TODO: update with translated string
+    error: 'Possible request to a loopback network address 127.0.0.1 in assets/app.js'
+  },
+  { type: 'link-local',
+    example_address: 'https://169.254.0.1',
```
seangoedecke: nit: worth testing that `http://` link-local addresses get a warning and an error?

lib/zendesk_apps_support/package.rb
```diff
@@ -88,6 +89,10 @@ def js_files
       @js_files ||= files.select { |f| f.to_s == 'app.js' || (f.to_s.start_with?('lib/') && f.to_s.end_with?('.js')) }
     end
 
+    def html_files
+      @html_files ||= files.select { |f| f =~ %r{.*\.html?$} }
```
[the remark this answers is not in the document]
fvzzy: Just consistency with the other file type checks tbh. Actually now that we've imported `mimemagic` in this repo (from the MIME types check) maybe we could use that here instead

lib/zendesk_apps_support/validations/requests.rb
fvzzy: ...good question. I think it's defaulting to the class `parse` method, though I'm not sure how exactly that works. I can look into this

spec/validations/requests_spec.rb
fvzzy: Nah, it wouldn't. The way ZAS is wired up, errors trump warnings and warnings are only displayed in the absence of errors. Not great since developers might have to do multiple validation passes but I think that's a change outside the scope of this PR

lib/zendesk_apps_support/validations/requests.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+require 'uri'
+require 'ipaddress'
+
+module IPAddress
+  class IPv4
+    # TODO: remove when/if a new version of https://github.com/ipaddress-gem/ipaddress is released
+    def link_local?
```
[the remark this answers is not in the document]
fvzzy: I'd rather not fork the repo just for this, but pinning to a Github SHA works

lib/zendesk_apps_support/package.rb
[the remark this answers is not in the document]
seangoedecke: `mimemagic` feels like overkill, but I bet there's something on the `File` class that would help

spec/validations/requests_spec.rb
seangoedecke: Ahh, makes sense.

lib/zendesk_apps_support/validations/requests.rb
seangoedecke: Ah, I see: https://github.com/ipaddress-gem/ipaddress/blob/a8eb06d418efc1b3bb036c849c95cf0ce4b7e18b/lib/ipaddress.rb#L276

TBH I have a slight preference for just `IPAddress::parse` - this kind of syntactic sugar makes my skin crawl.

fvzzy: Yep that's fair - I'll update this

spec/validations/requests_spec.rb
fvzzy: It _could_, but I think we sacrifice a fair bit of readability for the sake of condensing 4 cases into 2 here. On the other hand avoiding `shared_examples` all together pushes this out to a _minimum_ of 16 cases (i.e. if we only test these 4 http function call types and no others).

I actually think it's worth keeping these seperate. What do you think?

spec/validations/requests_spec.rb
fvzzy: Admittedly I think this is a bit gross as well - I scratched my head about this for a long time and eventually ended up here trying to make the assertion readable, e.g. `ZendeskAppsSupport::Validations::Requests to 127.0.0.0/8 range addresses behaves like a blocked loopback request and error inside XMLHttpRequest style request calls`. `it_behaves_like` is really awkward here.

I think your suggestion makes sense... I'll give that a go

lib/zendesk_apps_support/package.rb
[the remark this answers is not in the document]
fvzzy: Do you mind if I leave this for the time being and add a seperate story to tidy these up?

spec/validations/requests_spec.rb
fvzzy: Hmm actually this might be OK in conjunction with your suggestion below. I'll give this a go.

lib/zendesk_apps_support/package.rb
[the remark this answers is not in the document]
seangoedecke: Yeah, no worries - this is definitely just a nit.

lib/zendesk_apps_support/validations/requests.rb
```diff
@@ -0,0 +1,44 @@
...
+  module Validations
+    module Requests
+      class << self
+        HTTP_REQUEST_METHODS = %w[GET HEAD POST PUT DELETE CONNECT OPTIONS TRACE].freeze
+
+        REQUEST_CALLS = [
+          /\w+\.request\(['"](.*)['"]/i, # ZAF request
+          /\$\.(?:get|post|getJSON)\(['"](.*)['"]/i, # jQuery request
+          /\w+\.open\(['"](?:#{Regexp.union(HTTP_REQUEST_METHODS).source})['"],\s?['"](.*)['"]/i, # XMLHttpRequest
+          /fetch\(['"](.*)['"]/i # fetch
```
mtsolakiszen: Should these regex's be case insensitive? `/i`?

lib/zendesk_apps_support/validations/requests.rb
```diff
@@ -0,0 +1,44 @@
...
+
+module ZendeskAppsSupport
+  module Validations
+    module Requests
+      class << self
+        HTTP_REQUEST_METHODS = %w[GET HEAD POST PUT DELETE CONNECT OPTIONS TRACE].freeze
+
+        REQUEST_CALLS = [
+          /\w+\.request\(['"](.*)['"]/i, # ZAF request
+          /\$\.(?:get|post|getJSON)\(['"](.*)['"]/i, # jQuery request
```
mtsolakiszen: Should we also check for the`ajax` function?
Also worth checking for `jQuery.{function_name}` as well as `$.{function_name}`?

lib/zendesk_apps_support/validations/requests.rb
fvzzy: Yep, intentional (i.e. `$.GET()` is also valid)

lib/zendesk_apps_support/validations/requests.rb
fvzzy: Mm yep, good call. I'll update these

Gemfile.lock
```diff
@@ -1,3 +1,10 @@
+GIT
+  remote: https://github.com/ipaddress-gem/ipaddress
+  revision: a172928b3a29c9286800bf8e4d9a9f13f0cf738a
+  ref: a172928
+  specs:
+    ipaddress (0.8.3)
+
```
keitheous: A gentle reminder to remove this reference before merging.

fvzzy: This is intentionally pinned to a Github branch as the source has been updated to include a feature we require here (`.link_local`) but the gem was never released since the update. Only the Gemfile allows pointing to a specific branch/commit like this which is why it sits here and not in the gemspec.

I'll add a comment to clear this up

keitheous:
While in the process of QAing this, we realised that there is still more work to do as some as ZAS is not looking at the right files. `package.js_files` needs to be redefined and `source.rb` needs to be worked on. 

1) `package.js_files` is defined by a regex to look for all js files in app folder - with reference to v1's `app.js`
2) `source.rb` is used to check V1 app with reference to `lib/app.css` and `lib/app.js` - needs to re think this

I was able to manually trigger this and able to see where the warnings are suppose to indicate, so that means a step in the right direction.

Screen Shot 2019-07-02 at 4 24 52 pm

Effectively moving this back to In Progress.

Gemfile.lock
keitheous: I believe this is merged to master which is released as v0.8.3.

To investigate this I looked at the specific commit that you were referring to
https://github.com/ipaddress-gem/ipaddress/commit/a172928b3a29c9286800bf8e4d9a9f13f0cf738a

Then I looked for these exact changes in the current version. They are here
ipv6: https://github.com/ipaddress-gem/ipaddress/blob/master/lib/ipaddress/ipv6.rb#L426-L453,
ipv4: https://github.com/ipaddress-gem/ipaddress/blob/master/lib/ipaddress/ipv4.rb#L653-L666

Could I overlooked something in this investigation?

keitheous: After discussing in person, I now understand what Farhan meant.

Version 0.8.3 was publish 2 years ago, but this change with `link_local?` is recent and has not been merged. 

[This link](https://github.com/ipaddress-gem/ipaddress/compare/v0.8.3...a172928b3a29c9286800bf8e4d9a9f13f0cf738a) will indicate the time line and real difference.

keitheous:
Keith's TODO List
- [ ] Merge this branch with `package.js_files` redefined.
- [ ] Rebase this branch off master to fast forward branch
- [ ] QA again and merge this branch if pass

PR #233, opened 2019-07-03
Adding Sell product and Sell-related locations.

grzegorz-dyk:
These locations will be used firstly by Sell app market. Sell-specific
apps will be able to determine in which Sell location they will appear
at.

This is part of Sell's App Framework project.

### DIFF
--- a/lib/zendesk_apps_support/location.rb
+++ b/lib/zendesk_apps_support/location.rb
@@ -50,12 +50,16 @@ module ZendeskAppsSupport
                    product_code: Product::STANDALONE_CHAT.code, v2_only: false),
       Location.new(id: 14, name: 'background',
                    product_code: Product::CHAT.code),
-      Location.new(id: 15, name: 'deal_card', product_code: Product::SELL.code),
-      Location.new(id: 16, name: 'contact_card', product_code: Product::SELL.code),
-      Location.new(id: 17, name: 'lead_card', product_code: Product::SELL.code),
-      Location.new(id: 18, name: 'dashboard', product_code: Product::SELL.code),
-      Location.new(id: 19, name: 'background', product_code: Product::SELL.code),
-      Location.new(id: 20, name: 'modal', product_code: Product::SELL.code)
+      Location.new(id: 15, name: 'top_bar', product_code: Product::SELL.code),
+      Location.new(id: 16, name: 'nav_bar', product_code: Product::SELL.code),
+      Location.new(id: 17, name: 'deal_card', product_code: Product::SELL.code, collapsible: true),
+      Location.new(id: 18, name: 'contact_card', product_code: Product::SELL.code, collapsible: true),
+      Location.new(id: 19, name: 'lead_card', product_code: Product::SELL.code, collapsible: true),
+      Location.new(id: 20, name: 'background', product_code: Product::SELL.code),
+      Location.new(id: 21, name: 'modal', product_code: Product::SELL.code),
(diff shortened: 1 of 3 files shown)


lib/zendesk_apps_support/product.rb
```diff
@@ -19,11 +19,13 @@ def self.all
     PRODUCTS_AVAILABLE = [
       Product.new(code: 1, name: 'support', legacy_name: 'zendesk'),
       Product.new(code: 2, name: 'chat', legacy_name: 'zopim'),
-      Product.new(code: 3, name: 'standalone_chat', legacy_name: 'lotus_box')
+      Product.new(code: 3, name: 'standalone_chat', legacy_name: 'lotus_box'),
+      Product.new(code: 4, name: 'sell', legacy_name: 'base')
```
grzegorz-dyk: @anyone Do I need to populate these in the DB? If yes, how can I do it?

grzegorz-dyk: Also, not sure If Sell needs `legacy_name`? We did not appear under legacy name in Zendesk Suite...

bbrzoska: @grzegorz-dyk I see `legacy_name` used here: https://github.com/zendesk/zendesk_app_market/blob/9036852175a27ddb40a1d97d9b8861951053df18/app/models/location.rb#L19

we could use the .name if .legacy_name doesn't exist in `app_market` or simply define `sell` in both cases.

lib/zendesk_apps_support/location.rb
```diff
@@ -49,7 +49,17 @@ def self.all
       Location.new(id: 13, name: 'system_top_bar',
                    product_code: Product::STANDALONE_CHAT.code, v2_only: false),
       Location.new(id: 14, name: 'background',
-                   product_code: Product::CHAT.code)
+                   product_code: Product::CHAT.code),
+      Location.new(id: 15, name: 'top_bar', product_code: Product::SELL.code),
```
Mehul-Porwal: Is Sell really building so many locations?

grzegorz-dyk: @Mehul-Porwal We do have these in frontend. I was advised to add these typical ones (`nav_bar`, `top_bar` etc).

bbrzoska: @Mehul-Porwal we're launching with 4 location at start, but we'll be adding others listed sometime next year. If it's a relatively easy change to add new locations in the DB, we can start with 4 and take it from there.

Mehul-Porwal: You do not need to add it to DB... just adding here is all that is required. Maybe just add the 4 you need at the moment and add more when required.

bbrzoska: I see, thanks @Mehul-Porwal. Why is the DB mentioned here though?

grzegorz-dyk: @bbrzoska I asked whether I should add them to DB in other comment. It did not get attached to thread though :/

Mehul-Porwal: The ids defined here are used in DB, when someone installs an app for 1Location.new(id: 17, name: 'deal_card', product_code: Product::SELL.code, collapsible: true),` we will use the code `17` in DB for the app_location, which will then we used to link to the definition here. Which means once in production, you can never change the id codes.

bbrzoska: Ah, that makes total sense. Thanks for the explanation @Mehul-Porwal!

PR #234, opened 2019-07-04
[MPORT-498] Update Source Validation and Remove jshintrb

keitheous:
🐶 🐻 

cc: @zendesk/dingo @zendesk/wattle-dev 

### Description && Approach
Recently we worked on another security task that [checks for insecure and block-listed requests](https://github.com/zendesk/zendesk_apps_support/pull/230). To do this, we obtain all `package.html_templates` and `package.js_files` to check for lines that look like requests using regex.

During the QA process, we realised that `package.js_files` is defined inaccurately, probably based on how V1 apps were structured. **The first objective here is to change** `package.js_files` **to grab ALL javascript files in package.**

  Screen Shot 2019-07-04 at 12 30 14 pm

`package.js_files` is used in one other place, `source.rb`. Here, I identified that the main javascript source is identified to be [app.js](https://github.com/zendesk/zendesk_apps_support/blob/master/lib/zendesk_apps_support/validations/source.rb#L28). I realised, I cannot change the definition of `package.js_files` without updating `source.rb` ~otherwise I would punch my laptop~

Screen Shot 2019-07-04 at 5 24 07 pm

Running this validation check against a V2 app never worked because ZAS cannot locate this javascript source as defined above. These days we name them anything we want or `ticket_sidebar.js` or `background.js`. And yet we cannot abandon this class completely because this class exist for a reason.

This class has 3 responsibilities.
1) Block app upload with this validation error if an app is classified as marketingOnly, requirementsOnly or IframeOnly, AND contains source files.
  
    Screen Shot 2019-07-04 at 5 21 22 pm

2) Block app upload with this validation error if an app is NOT classified as marketingOnly, requirementsOnly or IframeOnly, AND it DOES NOT contain any source file (HTML, CSS and JS)
    
    Screen Shot 2019-07-04 at 5 21 15 pm

3) Run Linter against the javascript files if app is NOT classified as marketingOnly, requirementsOnly or IframeOnly.

I also realised that IframeOnly [here](https://github.com/zendesk/zendesk_apps_support/blob/master/lib/zendesk_apps_support/manifest.rb#L106) means any V2 apps. I initially thought it refers to an app with a manifest.location pointing to a server somewhere. These days, V2 apps can have multiple locations but have one location pointing to a server hosted else where. **Therefore, the check for IframeOnly should be removed from the class responsibilities.**

Next, do we really still need to run linter on javascript files? Developers allowed to have a different linters for their projects. So we **Remove Linter from the class responsibilities**.

So now we have a clearer picture of what our class should do.

**Block app upload with this validation error if an app is classified as marketingOnly or requirementsOnly, AND if it contains source files**

Screen Shot 2019-07-04 at 5 21 22 pm

### TODOs
- [ ] Check that regular uploads still work (private, special, normal v2 apps, v1 app?)
- [x] QA: Check that ZAT still works with this update
- [ ] QA: Check that ZAM still works with this update
- [ ] Merge this branch before merging [this](https://github.com/zendesk/zendesk_apps_support/pull/230)

### References
https://zendesk.atlassian.net/browse/MPORT-498

### Risk
- Medium: App upload fail in ZAM and ZAT. Apps deployed through Samson fail. Angry partners.


keitheous:
Development Test on ZAT
Screen Shot 2019-07-04 at 5 53 38 pm

lib/zendesk_apps_support/package.rb
```diff
@@ -94,7 +94,7 @@ def text_files
     end
 
     def js_files
-      @js_files ||= files.select { |f| f =~ %r{^(app|lib\/.*)\.js$} }
+      @js_files ||= files.select { |f| f =~ %r{^(.*)\.jsx?$} }
```
fvzzy: Probably don't need the capturing group for `(.*)` here

lib/zendesk_apps_support/validations/source.rb
```diff
@@ -1,68 +1,23 @@
...
-            errors << error unless error.nil?
-          end
+        def contain_source_files?(package)
+          package.js_files.any? || package.template_files.any? || !package.app_css.empty?
         end
 
-        def linter
-          Jshintrb::Lint.new(LINTER_OPTIONS)
+        def no_source_required_app?(manifest)
+          manifest.requirements_only? || manifest.marketing_only?
```
seangoedecke: What about `.iframe_only?` - do we still need that here?

zendesk_apps_support.gemspec
```diff
@@ -19,7 +19,6 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'json'
   s.add_runtime_dependency 'image_size'
   s.add_runtime_dependency 'erubis'
-  s.add_runtime_dependency 'jshintrb', '~> 0.3.0'
```
seangoedecke: Nice, long past time we got rid of jshint

keitheous: :laugh

lib/zendesk_apps_support/validations/source.rb
keitheous: The definition of`.iframe_only?` here refers to any [V2 app](https://github.com/zendesk/zendesk_apps_support/blob/master/lib/zendesk_apps_support/manifest.rb#L106).

Any V2 app would typically have source files such as HTML, CSS and JS. Therefore it wouldn't be classified under `no_source_required_app?`

Putting them together, we certainly do not want to block app upload with a validation because they package.manifest.is_a_v2_app && contain_source_files?(package).

Therefore I do not think we still need that there. What do you think?

lib/zendesk_apps_support/package.rb
keitheous: As discussed, I will be changing this to `%r{^.*\.jsx?$}`.

lib/zendesk_apps_support/validations/source.rb
seangoedecke: I'm pretty sure an iframe-only V2 app can just link out to some externally-hosted assets in the manifest, rather than having CSS/JS in the zip itself.

keitheous: Oh yes I agree Sean, a V2 app can have an internal source, ie. `assets/ticket_sidebar.js` and `assets/ticket_sidebar.css` AND a V2 app can have a location pointing to a server hosted somewhere else.

Hence I have removed this check in the PR. We do not want to block app upload with any validation error if the app is of both types. Let's discuss this further in person :)

keitheous: As discussed in person, I will rename this method to be `app_doesnt_require_source?` to avoid confusion. Thank you for the suggestion!

fvzzy:
@keitheous Would you mind rewording the title for this PR? I'm not sure it's accurate anymore, especially given a large part of the diff here is removing JSHint

keitheous:
@fvzzy What would you rename the title for this PR?
I think update is still appropriate because when we moved from V1 to V2 - alot of these checks were valid for a V1 app but no longer valid for V2. The title should encapsulate removing JSHint as well, checking that users do not do not include source files if they are uploading a requirementsOnly/marketingOnly app.

spec/validations/source_spec.rb
```diff
@@ -1,139 +1,47 @@
...
-      before do
-        allow(package).to receive(:js_files) { [double('AppFile', relative_path: 'lib/slapp.js')] }
-      end
-
-      it 'should have an error' do
-        errors = ZendeskAppsSupport::Validations::Source.call(package)
-        expect(errors.first.key).to eql :no_code_for_ifo_notemplate
-      end
+    it 'returns nothing if package has no source files' do
+      expect(validation_error).not_to receive(:new).with(:no_code_for_ifo_notemplate)
```
fvzzy: Would dropping the `.with(:no_code_for_ifo_notemplate)` condition here make this a better test? You'd never expect an error (of any kind) in this case right?

Also I found this a little odd to read with the expectation the only thing inside the `it` block. It's a bit redundant, but for readability maybe it's worth explicitly assigning `package.js_files = []` (or `package.files = []`)?

spec/validations/source_spec.rb
```diff
@@ -1,139 +1,47 @@
...
+      expect(validation_error).to receive(:new).with(:no_code_for_ifo_notemplate)
     end
-  end
 
-  context 'when not requirements only' do
-    it 'should have an error when app.js is missing' do
-      errors = ZendeskAppsSupport::Validations::Source.call(package)
-      expect(errors.first.to_s).to eql 'Could not find app.js'
+    it 'returns nothing if package has no source files' do
+      expect(validation_error).not_to receive(:new).with(:no_code_for_ifo_notemplate)
```
fvzzy: Same suggestion as https://github.com/zendesk/zendesk_apps_support/pull/234/files#r300911795

fvzzy: Actually this test is duplicate now. Perhaps it can be removed? Otherwise `package.template_files = []` first to differentiate it?

fvzzy:
+1 on the super detailed description, but for easier scanning maybe a few dot points to summarise the major changes in this PR.

Left a few really minor comments, otherwise 👍

spec/validations/source_spec.rb
keitheous: As discussed in person, js_files and template_files in package is defined on [this line](https://github.com/zendesk/zendesk_apps_support/blob/ce268a70cbefba0276d7dcaf7558aa5b37ad114b/spec/validations/source_spec.rb#L8)

```
line 8
let(:package) { double('Package', js_files: [], template_files: [], app_css: '') }
```

spec/validations/source_spec.rb
keitheous: Yeap let me address that.

PR #239, opened 2019-07-21
[MPORT-502] Rewrite Requests validation class

keitheous:
@zendesk/dingo @zendesk/vegemite @zendesk/wattle 

### Description
This rewrite is necessary because of there are some edge cases that were not considered, which resulted in App uploads failing in production.

**The initial approach** includes **searching for Request Methods** and **evaluating these addresses**.

**Problems discovered:**  
1) Regex grabs everything after the request methods including `.then(..).catch(...)`. 
  ```
  /api/v2/channels/voice/stats/current_queue_activity.json").then(function requestDone(data)
{cqa=data.current_queue_activity,resolve()})})}function getAgentActivity(){return
 agents.online=0,agents.available=0,new Promise(function(resolve,reject)
{client.request("/api/v2/channels/voice/stats/agents_activity.json").then(
function requestDone(data)
  ...
  ...
  URI(request.captures[0]) -> URI::InvalidURIError: bad URI(is not URI?)
  ```
  2) `package.read` converts everything to **double quoted strings**, retaining the `\"` and `\t` as well, therefore regexing off this wouldn't work effectively.
  3) There is always the possibility we can miss the address when a developer does this.

  ```
  var url = 'http://www.google.com/abc'
  client.request(url).then(successHandler).catch(failHandler)
  ```

  4) Also as we loop through each files, and with every file loop, we loop through the four request methods regex. Don't know what is the O notation on that. This can be reduced by a simple Regex union.

**In this new approach,** we scan for everything that looks like a url or ip address. We also `file.read.gsub(/\"/, '')`.

### Approach
1) When we detect `http` urls, we call this out in a warning.

2) When we detect any IP address, we determine if they are either a loopback, link_local or private.

Screen Shot 2019-07-21 at 2 34 59 pm

Screen Shot 2019-07-21 at 2 35 19 pm

### TODOs:
- [x] Update Specs
- [x] Use localization for `.join(', ')` - `txt.apps.admin.error.app_build.listing_comma`
- [x] Create localization branch for new string

### References
- https://zendesk.atlassian.net/browse/MPORT-502

### Risks
[low] App upload fails when validating app

### DIFF
--- a/lib/zendesk_apps_support/validations/requests.rb
+++ b/lib/zendesk_apps_support/validations/requests.rb
@@ -1,56 +1,68 @@
 # frozen_string_literal: true
 
-require 'uri'
 require 'ipaddress_2'
+require 'uri'
 
 module ZendeskAppsSupport
   module Validations
     module Requests
       class << self
-        HTTP_REQUEST_METHODS = %w[GET HEAD POST PUT DELETE CONNECT OPTIONS TRACE].freeze
-
-        REQUEST_CALLS = [
-          /\w+\.request\(['"](.*)['"]/i, # ZAF request
-          /(?:\$|jQuery)\.(?:ajax|get|post|getJSON)\(['"](.*)['"]/i, # jQuery request
-          /\w+\.open\(['"](?:#{Regexp.union(HTTP_REQUEST_METHODS).source})['"],\s?['"](.*)['"]/i, # XMLHttpRequest
-          /fetch\(['"](.*)['"]/i # fetch
-        ].freeze
+        IP_ADDRESS = /\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}/
 
         def call(package)
           errors = []
           files = package.js_files + package.html_files
 
           files.each do |file|
-            contents = file.read
-            REQUEST_CALLS.each do |request_pattern|
-              request = contents.match(request_pattern)
-              next unless request
-              uri = URI(request.captures[0])
-              if uri.scheme == 'http'
-                package.warnings << I18n.t('txt.apps.admin.warning.app_build.insecure_http_request',
-                                           uri: uri,
-                                           file: file.relative_path)
-              end
+            file_content = file.read
 
-              next unless IPAddress.valid? uri.host
-              ip = IPAddress.parse uri.host
+            http_protocol_urls = find_address_containing_http(file_content)
+            if http_protocol_urls.any?
+              package.warnings << I18n.t(
+                'txt.apps.admin.warning.app_build.insecure_http_request',
+                uri: http_protocol_urls.join(I18n.t('txt.apps.admin.error.app_build.listing_comma')),
+                file: file.relative_path
+              )
+            end
 
-              blocked_ip_type = if ip.private?
-                                  I18n.t('txt.apps.admin.error.app_build.blocked_request_private')
-                                elsif ip.loopback?
-                                  I18n.t('txt.apps.admin.error.app_build.blocked_request_loopback')
-                                elsif ip.link_local?
-                                  I18n.t('txt.apps.admin.error.app_build.blocked_request_link_local')
-                                end
-
-              next unless blocked_ip_type
-              errors << I18n.t('txt.apps.admin.error.app_build.blocked_request',
-                               type: blocked_ip_type,
-                               uri: uri.host,
-                               file: file.relative_path)
+            ip_addresses = file_content.scan(IP_ADDRESS)
+            if ip_addresses.any?
+              errors << blocked_ips_validation(file.relative_path, ip_addresses)
             end
           end
+
           errors
+        end
+
+        private
+
+        def blocked_ips_validation(file_path, ip_addresses)
+          ip_addresses.each_with_object([]) do |ip_address, error_messages|
+            blocked_type = blocked_ip_type(ip_address)
+            next unless blocked_type
+
+            error_messages << I18n.t(
+              'txt.apps.admin.error.app_build.blocked_request',
+              type: blocked_type,
+              uri:  ip_address,
+              file: file_path
+            )
+          end
+        end
+
+        def blocked_ip_type(ip_address)
+          block_type =
+            case IPAddress.parse(ip_address)
+            when proc(&:private?) then 'private'
+            when proc(&:loopback?) then 'loopback'
+            when proc(&:link_local?) then 'link_local'
+            end
+
+          block_type && I18n.t("txt.apps.admin.error.app_build.blocked_request_#{block_type}")
+        end
+
+        def find_address_containing_http(file_content)
+          file_content.scan(URI.regexp(['http'])).map(&:compact).map(&:last)
         end
       end
     end
--- a/spec/validations/requests_spec.rb
+++ b/spec/validations/requests_spec.rb
@@ -2,75 +2,52 @@
 
 require 'spec_helper'
 
-def request_function(style, address)
-  case style
-  when 'ZAF'
-    "function request() { client.request('#{address}') }"
-  when 'jQuery'
-    "function request() { jQuery.get('#{address}') }"
-  when 'jQuery$'
-    "function request() { $.ajax('#{address}') }"
-  when 'XMLHttpRequest'
-    "function request() { xhr.open('get', '#{address}') }"
-  when 'fetch'
-    "function request() { fetch('#{address}') }"
-  end
-end
-
-request_function_styles = %w[ZAF jQuery jQuery$ XMLHttpRequest fetch]
-
-shared_examples 'an insecure request' do |file_path, function_style|
-  address = 'http://foo.com'
-  let(:markup) { request_function(function_style, address) }
-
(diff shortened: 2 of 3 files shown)


lib/zendesk_apps_support/validations/requests.rb
```diff
@@ -1,57 +1,108 @@
...
+                uri: http_protocol_urls.join(', '),
+                file: file.relative_path
+              )
+            end
+
+            url_addresses = search_in_file(COMMON_URLS, file_content)
+            if url_addresses.any?
+              unrecognised_urls = find_unrecognised_urls(url_addresses)
+              if unrecognised_urls.any?
+                package.warnings << unrecognised_urls_warning(file.relative_path, unrecognised_urls)
```
seangoedecke: How come we're adding this new validation? Even if we do want to start warning on this, I wonder if it should be split out into a separate PR/story so the bugfixes and refactoring here can be reviewed independently

keitheous: That is an excellent idea! I will put this in a separate PR.

As discussed in person, the reason why I put it in here was because I saw this in Kari's work in AppsApproval. 

Running this by Chris, he mentioned that this can be valuable in the apps upload process in ZAM for Eric. Therefore, I will think about how I can move this there in a seperate ticket.

lib/zendesk_apps_support/validations/requests.rb
```diff
@@ -1,57 +1,70 @@
...
+            http_protocol_urls = search_in_file(HTTP_URI, file_content)
+            if http_protocol_urls.any?
+              package.warnings << I18n.t(
+                'txt.apps.admin.warning.app_build.insecure_http_request',
+                uri: http_protocol_urls.join(I18n.t('txt.apps.admin.error.app_build.listing_comma')),
+                file: file.relative_path
+              )
+            end
+
+            ip_addresses = search_in_file(IP_ADDRESS, file_content)
```
seangoedecke: We're iterating over each file twice - could potentially speed that up by checking each line for http urls and ips in one go

lib/zendesk_apps_support/validations/requests.rb
```diff
@@ -1,57 +1,70 @@
...
+              uri:  ip_address,
+              file: file_path
+            )
+          end
+        end
+
+        def blocked_ip_type(ip_address)
+          block_type =
+            case IPAddress.parse(ip_address)
+            when proc(&:private?) then 'private'
```
seangoedecke: Pretty fancy stuff

lib/zendesk_apps_support/validations/requests.rb
```diff
@@ -1,57 +1,70 @@
...
-                                elsif ip.link_local?
-                                  I18n.t('txt.apps.admin.error.app_build.blocked_request_link_local')
-                                end
-
-              next unless blocked_ip_type
-              errors << I18n.t('txt.apps.admin.error.app_build.blocked_request',
-                               type: blocked_ip_type,
-                               uri: uri.host,
-                               file: file.relative_path)
+            file_content = file.read.split(' ').map { |word| word.gsub(/[\"\']/, '') }
```
seangoedecke: How come we're stripping out quotes here?

lib/zendesk_apps_support/validations/requests.rb
```diff
@@ -1,57 +1,70 @@
...
-require 'uri'
 require 'ipaddress_2'
 
 module ZendeskAppsSupport
   module Validations
     module Requests
       class << self
-        HTTP_REQUEST_METHODS = %w[GET HEAD POST PUT DELETE CONNECT OPTIONS TRACE].freeze
-
-        REQUEST_CALLS = [
```
seangoedecke: I agree with stripping out the JS-request-parsing stuff in favour of just checking for the presence of an internal IP or http address. It'd be nice if we could do this, but it's really hard to catch all the edge cases with just regex.

lib/zendesk_apps_support/validations/requests.rb
seangoedecke: (I'm probably happy with doing it twice, by the way, just pointing it out in case it does end up being a performance problem for large apps. Have you tested it with a big app zip?)

lib/zendesk_apps_support/validations/requests.rb
keitheous: Thank you, Sean.

lib/zendesk_apps_support/validations/requests.rb
keitheous: Thank you, Sean.

lib/zendesk_apps_support/validations/requests.rb
keitheous: Good point. Now that I am no longer looking for URIs, I can probably do without this. Let me investigate and get back to you shortly. :)

lib/zendesk_apps_support/validations/requests.rb
keitheous: 👍 

It is true. We could do everything in one iteration. But I might stick with the two for now :)

lib/zendesk_apps_support/validations/requests.rb
keitheous: Thank you for the suggestion on reviewing the files as a string (as a whole), @seangoedecke .

I can definitely see how it is concerning if we are scanning through a big file. We are also no longer gsubbing the strings.

spec/validations/requests_spec.rb
```diff
@@ -2,75 +2,52 @@
...
+  before { allow(package).to receive(:js_files) { [app_file] } }
 
-  it "and raise a warning inside #{function_style} style requests" do
-    errors = subject.call(package)
-    expect(package.warnings[0]).to include('insecure HTTP request', address, file_path)
-    expect(errors).to be_empty
-  end
-end
+  context 'http protocols check' do
+    it 'returns no warnings when files do not contain http url' do
```
JeanMarcGoepfert: Might be more accurate to say `returns no warnings when files that contain http**s** urls`

spec/validations/requests_spec.rb
```diff
@@ -2,75 +2,52 @@
...
-  }
-}
+      subject.call(package)
+      expect(package.warnings).to be_empty
+    end
 
-shared_examples 'a blocked ip' do |file_path, function_style, ip_type, ip|
-  let(:markup) { request_function(function_style, "https://#{ip}") }
+    it 'returns warning with request information when files contain http url' do
+      allow(app_file).to receive(:read) { "client.instance(\"http:/foo-bar.com\");\r\n\t" }
```
JeanMarcGoepfert: There should be 2 `/` characters before the domain.

lib/zendesk_apps_support/validations/requests.rb
```diff
@@ -1,57 +1,69 @@
...
-          /\w+\.request\(['"](.*)['"]/i, # ZAF request
-          /(?:\$|jQuery)\.(?:ajax|get|post|getJSON)\(['"](.*)['"]/i, # jQuery request
-          /\w+\.open\(['"](?:#{Regexp.union(HTTP_REQUEST_METHODS).source})['"],\s?['"](.*)['"]/i, # XMLHttpRequest
-          /fetch\(['"](.*)['"]/i # fetch
-        ].freeze
+        HTTP_URI = /[\'\"](http[^s].*?)[\'\"]/i
```
JeanMarcGoepfert: Are the `[\'\"]` blocks needed here? 

Also, might be better to just use the [URI::regexp](https://apidock.com/ruby/URI/regexp/class) for this...

seangoedecke: +1 on using URI::regexp

JeanMarcGoepfert:
Looks good @keitheous, just left some minor comments.

As the new guy I'd be keen to know what our long term thinking around security is for this stuff. Anyone could get around what we're doing now by doing something as simple as `var ip = "127" + ".0.0.1" in their code.

lib/zendesk_apps_support/validations/requests.rb
keitheous: Yeah this will work

```
(byebug) file_content.scan(URI.regexp(["http"]))
[["http", "hello.com'", nil, nil, nil, nil, nil, nil, nil], ["http", "wikipedia.org'", nil, nil, nil, nil, nil, nil, nil], ["http", nil, nil, "www.foo-bar.com", nil, nil, nil, nil, nil]]

(byebug) file_content.scan(URI.regexp(["http"])).map(&:compact).map(&:last)
["hello.com'", "wikipedia.org'", "www.foo-bar.com"]
```

`.match` only returns the first one it finds
```
(byebug) file_content.match(URI.regexp(['http']))
#<MatchData "http:hello.com'" 1:"http" 2:"hello.com'" 3:nil 4:nil 5:nil 6:nil 7:nil 8:nil 9:nil>
```

keitheous: Thank you for the suggestion on using `URI.regexp(['http'])`, @JeanMarcGoepfert .

PR #245, opened 2019-08-07
[MPORT-518] Give more information in secrets warnings

seangoedecke:
This PR adds a string to the secrets warning:
```
sgoedecke:private_app/ $ zat_dev validate                                                                                         [11:26:06]
     warning  Possible secrets found in assets/foo.js ('password...'), assets/shared.js ('secret...'). Consider reviewing the contents of these files before submitting your app.
    validate  OK
```

It also removes some keywords from the secret validation ('username' and 'pass') which I think are too liable to cause false positives. It's OK to have a username variable, and 'pass' is a substring of 'passed', which can come up a lot when writing code.

https://zendesk.atlassian.net/browse/MPORT-518

### DIFF
--- a/lib/zendesk_apps_support/validations/secrets.rb
+++ b/lib/zendesk_apps_support/validations/secrets.rb
@@ -60,7 +60,8 @@ module ZendeskAppsSupport
                                          secret_type: secret_type)
             end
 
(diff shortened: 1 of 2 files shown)


seangoedecke:
cc @zendesk/dingo

PR #248, opened 2019-08-08
[MPORT-528] Ignores anything that is not an IP Address

fbvilela:
## Description

Fixes bug found in production

### JIRA
https://zendesk.atlassian.net/browse/MPORT-528

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -27,7 +27,7 @@ GEM
       i18n (~> 0.5)
     ffi (1.9.25)
     i18n (0.7.0)
-    image_size (2.0.2)
+    image_size (2.0.1)
     ipaddress_2 (0.13.0)
     json (2.2.0)
     loofah (2.2.3)
@@ -91,4 +91,4 @@ DEPENDENCIES
   zendesk_apps_support!
 
 BUNDLED WITH
-   2.0.1
+   1.16.3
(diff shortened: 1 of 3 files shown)


spec/validations/requests_spec.rb
```diff
@@ -40,6 +40,11 @@
       expect(validation_error_class).to_not receive(:new)
     end
 
+    it 'ignores numbers that are invalid IP Adresses' do
+      allow(app_file).to receive(:read) { script_containing_ip('857.384.857.857') }
+      # would throw an exception and this test would fail prior to the fix.
```
seangoedecke: IMO we don't need this comment

keitheous:
Rubocop errors:
- Do not suppress exceptions.
- Useless assignment to variable - "error".

This would work
```
        def blocked_ip_type(ip_address)
          block_type =
            case IPAddress.parse(ip_address)
            when proc(&:private?) then 'private'
            when proc(&:loopback?) then 'loopback'
            when proc(&:link_local?) then 'link_local'
            end

          block_type && I18n.t("txt.apps.admin.error.app_build.blocked_request_#{block_type}")
        rescue ArgumentError
          nil # Ignore numbers which are not an IP address
        end
```

Gemfile.lock
```diff
@@ -91,4 +91,4 @@ DEPENDENCIES
   zendesk_apps_support!
 
 BUNDLED WITH
-   2.0.1
+   1.16.3
```
keitheous: https://github.com/zendesk/zendesk_apps_support/commit/44008a006cfbfc306462d9684c0372668ff30b58

PR #249, opened 2019-08-27
[SAF-853] Divided contact_card location to person_card and company_card

rprzebieglec:
We'd like to be more specific when saying whether app is displayed on person or on company contact. 

This PR removed previous `contact_card` and adds `person_card` and `company_card`.

@Mehul-Porwal Is this kind of change safe? I assume there's 0 apps with previously defined IDs, so that shouldn't be a problem, right?

This PR also adds a new location `editor` - this is simmilar to `ticket_editor` location in Support. We already have a potential app ideas, and since I'm messing with locations I added it now :)


rprzebieglec:
@bbrzoska that is a legit concern - will confirm with product

rprzebieglec:
OK, So I also split editor location into three distinct ones.

@Mehul-Porwal What should be the next steps with this PR?

rprzebieglec:
cc: @zendesk/wattle  @zendesk/vegemite

PR #252, opened 2019-09-11
[MPORT-550] Lockdown Bundler version

keitheous:
**Description:**
In the past, Gemfile.lock's `BUNDLE WITH` gets bumped by accident, mostly from gem deploy commands or rake bump commands. If this version goes above 2.0.0, our travis build will fail because we support stone-age ruby.

Jira: https://zendesk.atlassian.net/browse/MPORT-550

**Risk:** 
- Low: Developers can't bundle ZAS dependencies

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,10 +9,16 @@ This repo is owned and maintained by the Zendesk Apps team. You can reach us on 
 ## Getting Started
 When you want to help **develop** this tool, you will need to clone this repo.
 
+Since ZAS is used in ZAT, which supports Ruby 2.1, the Bundler Gem Version 1.17.3 is required for bundling dependencies. If you are working with a version that is higher than `1.17.3`, you will need to downgrade it.
+```
+  $ gem install bundler --version 1.17.3
+  $ bundle install
+```
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -9,10 +9,16 @@ This repo is owned and maintained by the Zendesk Apps team. You can reach us on
 ## Getting Started
 When you want to help **develop** this tool, you will need to clone this repo.
 
+The Bundler Gem Version 1.17.3 is required for bundling dependencies. If you are working with a version that is higher than `1.17.3`, you will need to downgrade it to work on ZAS.
```
seangoedecke: Let's give a reason here, e.g. "Since ZAS is used in ZAT, which supports Ruby 2.1, the Bundler Gem Version...."

keitheous: Oh that sounds GOOD!

keitheous: This will now be
```
Since ZAS is used in ZAT, which supports Ruby 2.1, the Bundler Gem Version 1.17.3 is required for bundling dependencies. If you are working with a version that is higher than `1.17.3`, you will need to downgrade it.
```

PR #253, opened 2019-10-10
[MPORT-580] Change blockers to warnings for private apps IP requests

keitheous:
cc: @zendesk/dingo @zendesk/wattle-dev  

[Private flag in app manifest is true by default. If a developer wishes to publish an app to the marketplace, this flag is set to false.](https://developer.zendesk.com/apps/docs/developer-guide/manifest#private)

In this implementation, when an app is private, these strings will show up as warnings instead. 

Jira: https://zendesk.atlassian.net/browse/MPORT-580

Risk: Low - App upload is blocked or fails

### DIFF
--- a/lib/zendesk_apps_support/validations/requests.rb
+++ b/lib/zendesk_apps_support/validations/requests.rb
@@ -12,45 +12,70 @@ module ZendeskAppsSupport
         def call(package)
           errors = []
           files = package.js_files + package.html_files
+          private_app = package.manifest.private?
 
           files.each do |file|
             file_content = file.read
 
             http_protocol_urls = find_address_containing_http(file_content)
             if http_protocol_urls.any?
-              package.warnings << I18n.t(
-                'txt.apps.admin.warning.app_build.insecure_http_request',
-                uri: http_protocol_urls.join(I18n.t('txt.apps.admin.error.app_build.listing_comma')),
-                file: file.relative_path
+              package.warnings << insecure_http_requests_warning(
+                http_protocol_urls,
+                file.relative_path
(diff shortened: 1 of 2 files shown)


lib/zendesk_apps_support/validations/requests.rb
```diff
@@ -12,45 +12,78 @@ class << self
...
+            uri: http_protocol_urls,
+            file: relative_path
+          )
+        end
+
+        def block_request_warning(string_params)
+          I18n.t('txt.apps.admin.error.app_build.blocked_request', string_params)
+        end
+
+        def validation_error(string_params)
```
seangoedecke: I don't think we need this new method, given it's only used in one place

lib/zendesk_apps_support/validations/requests.rb
```diff
@@ -12,45 +12,78 @@ class << self
...
+          )
+
+          I18n.t(
+            'txt.apps.admin.warning.app_build.insecure_http_request',
+            uri: http_protocol_urls,
+            file: relative_path
+          )
+        end
+
+        def block_request_warning(string_params)
```
seangoedecke: Likewise I think this can be inlined into the function

keitheous: 👍

lib/zendesk_apps_support/validations/requests.rb
keitheous: 👍

keitheous:
Thank you, Sean. Appreciate it! :)

PR #256, opened 2019-10-23
[MPORT-600] Add validation for custom object types and relationships

seangoedecke:
Adds validation for custom object requirements, against this schema:
```
{
  custom_objects: {
    custom_object_types: [ { key: 'dog', schema: {...}} ],
    custom_object_relationship_types: [ { key: 'dogs-join-cats', source: 'dog', target: 'cat' } ]
  }
}
```

Output:
```
sgoedecke@-:~/Code/zendesk/zendesk_apps_tools|master⚡
⇒  bundle exec zat validate --path=/Users/sgoedecke/Desktop/apps/private_app_with_custom_object_types
    validate  Missing required fields in requirements.json: "key" is required in "custom_object_types"
    validate  Missing required fields in requirements.json: "schema" is required in "custom_object_types"

sgoedecke@-:~/Code/zendesk/zendesk_apps_tools|master⚡
⇒ bundle exec zat validate --path=/Users/sgoedecke/Desktop/apps/private_app_with_custom_object_types
    validate  requirements.json contains an invalid type: custom_object_types

sgoedecke@-:~/Code/zendesk/zendesk_apps_tools|master⚡
⇒ bundle exec zat validate --path=/Users/sgoedecke/Desktop/apps/private_app_with_custom_object_types
    validate  Custom objects schema contains an invalid key: asdigfasdf

```

### DIFF
--- a/lib/zendesk_apps_support/validations/requirements.rb
+++ b/lib/zendesk_apps_support/validations/requirements.rb
@@ -126,7 +126,7 @@ module ZendeskAppsSupport
         def validate_custom_objects_keys(keys, expected_keys, identifier, errors = [])
           invalid_keys = keys - expected_keys
           unless invalid_keys.empty?
-            errors << ValidationError.new(:invalid_cr_schema_keys, # TODO: update error translations
+            errors << ValidationError.new(:invalid_custom_objects_schema_keys,
                                           invalid_keys: invalid_keys.join(', '),
                                           count: invalid_keys.length)
           end
--- a/lib/zendesk_apps_support/validations/requirements.rb
+++ b/lib/zendesk_apps_support/validations/requirements.rb
@@ -136,12 +136,6 @@ module ZendeskAppsSupport
         end
 
         def validate_custom_objects_keys(keys, expected_keys, identifier, errors = [])
-          invalid_keys = keys - expected_keys
-          unless invalid_keys.empty?
-            errors << ValidationError.new(:invalid_custom_objects_schema_keys,
-                                          invalid_keys: invalid_keys.join(', '),
-                                          count: invalid_keys.length)
-          end
           missing_keys = expected_keys - keys
           missing_keys.each do |key|
             errors << ValidationError.new(:missing_required_fields,
--- a/lib/zendesk_apps_support/validations/requirements.rb
+++ b/lib/zendesk_apps_support/validations/requirements.rb
@@ -26,7 +26,7 @@ module ZendeskAppsSupport
             errors << excessive_requirements(requirements)
             errors << invalid_channel_integrations(requirements)
             errors << invalid_custom_fields(requirements)
-            errors << invalid_custom_resources_schema(requirements)
+            errors << invalid_custom_objects(requirements)
             errors << missing_required_fields(requirements)
             errors.flatten!
             errors.compact!
@@ -44,7 +44,7 @@ module ZendeskAppsSupport
         def missing_required_fields(requirements)
           [].tap do |errors|
             requirements.each do |requirement_type, requirement|
-              next if %w[channel_integrations custom_resources_schema].include? requirement_type
+              next if %w[channel_integrations custom_objects].include? requirement_type
               requirement.each do |identifier, fields|
                 next if fields.include? 'title'
                 errors << ValidationError.new(:missing_required_fields,
@@ -92,23 +92,22 @@ module ZendeskAppsSupport
           end
         end
 
-        def invalid_custom_resources_schema(requirements)
-          custom_resources_schema = requirements['custom_resources_schema']
-          return unless custom_resources_schema
-          valid_schema_keys = %w[resource_types relationship_types]
-          [].tap do |errors|
-            invalid_keys = custom_resources_schema.keys - valid_schema_keys
-            unless invalid_keys.empty?
-              errors << ValidationError.new(:invalid_cr_schema_keys,
-                                            invalid_keys: invalid_keys.join(', '),
-                                            count: invalid_keys.length)
-            end
+        def invalid_custom_objects(requirements)
+          custom_objects = requirements['custom_objects']
+          return unless custom_objects
 
-            valid_schema_keys.each do |required_key|
-              next if custom_resources_schema.keys.include? required_key
-              errors << ValidationError.new(:missing_required_fields,
-                                            field: required_key,
-                                            identifier: 'custom_resources_schema')
+          valid_schema = {
+            'custom_object_types' => ['key', 'schema'],
+            'custom_object_relationship_types' => ['key', 'source', 'target']
+          }
+
+          [].tap do |errors|
+            validate_custom_objects_keys(custom_objects.keys, valid_schema.keys, 'custom_objects', errors)
+
+            valid_schema.keys.each do |requirement_type|
+              (custom_objects[requirement_type] || []).each do |requirement|
+                validate_custom_objects_keys(requirement.keys, valid_schema[requirement_type], requirement_type, errors)
+              end
             end
           end
         end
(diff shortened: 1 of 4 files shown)


seangoedecke:
@zendesk/dingo Ready for review

lib/zendesk_apps_support/validations/requirements.rb
```diff
@@ -4,6 +4,7 @@ module ZendeskAppsSupport
   module Validations
     module Requirements
       MAX_REQUIREMENTS = 5000
+      MAX_CUSTOM_OBJECTS_REQUIREMENTS = 10
```
seangoedecke: this is just a placeholder number for now

lib/zendesk_apps_support/validations/requirements.rb
```diff
@@ -92,23 +106,22 @@ def invalid_channel_integrations(requirements)
...
-        def invalid_custom_resources_schema(requirements)
-          custom_resources_schema = requirements['custom_resources_schema']
-          return unless custom_resources_schema
-          valid_schema_keys = %w[resource_types relationship_types]
+        def invalid_custom_objects(requirements)
+          custom_objects = requirements['custom_objects']
+          return unless custom_objects
+
+          valid_schema = {
+            'custom_object_types' => %w[key schema],
```
mtsolakiszen: `end_users_can_read` is an optional writable key

seangoedecke: Thanks, good catch! I've removed the check for extra keys.

lib/zendesk_apps_support/validations/requirements.rb
```diff
@@ -56,10 +58,22 @@ def missing_required_fields(requirements)
...
-          count = requirements.values.map(&:values).flatten.size
+          count = requirements.values.map do |req|
+            req.is_a?(Hash) ? req.values : req
+          end.flatten.size
           ValidationError.new(:excessive_requirements, max: MAX_REQUIREMENTS, count: count) if count > MAX_REQUIREMENTS
         end
 
+        def excessive_custom_objects_requirements(requirements)
+          custom_objects = requirements['custom_objects']
+          return if custom_objects.nil?
```
keitheous: Hello 👋 

How do you feel about changing `return if custom_objects.nil?` to `return unless custom_objects`, for consistency with line 111 in this file.

Also, I am curious what happens if someone includes the extreme edge case value `false` here,
i.e `requirements['custom_objects'] = false`. But perhaps this is not important. What do you think?

lib/zendesk_apps_support/validations/requirements.rb
```diff
@@ -121,6 +134,15 @@ def invalid_requirements_types(requirements)
...
           end
         end
+
+        def validate_custom_objects_keys(keys, expected_keys, identifier, errors = [])
+          missing_keys = expected_keys - keys
+          missing_keys.each do |key|
+            errors << ValidationError.new(:missing_required_fields,
+                                          field: key,
+                                          identifier: identifier)
+          end
```
keitheous: I don't think `errors` is being returned here.

Screen Shot 2019-11-04 at 11 08 58 am

keitheous: I can suggest another approach, which is to either use a dot inject/each with object do, append, and flatten. Something like this. 

Screen Shot 2019-11-04 at 11 18 19 am

spec/validations/requirements_spec.rb
```diff
@@ -194,48 +223,54 @@
...
+          'custom_objects' => { 'custom_object_types' => [{}], 'custom_object_relationship_types' => [] }
+        )
+      end
 
-    it 'creates an error' do
-      expect(errors.first.key).to eq(:invalid_cr_schema_keys)
-      expect(errors.first.data).to eq(invalid_keys: 'resources, relationships', count: 2)
+      it 'creates an error' do
+        expect(errors.first.key).to eq(:missing_required_fields)
+        expect(errors.first.data).to eq(field: 'key', identifier: 'custom_object_types')
```
keitheous: Instead of `errors.first.data` or `errors.first.key`. Would it be better to test that
ValidationError.new to be called with`:missing_required_fields , field: ..., identifier: ...`

```
it 'creates an error' do
  expect(ValidationError).to receive(:new).with(:missing_required_fields , field: 'key', identifier: 'custom_object_types')
end
```

lib/zendesk_apps_support/validations/requirements.rb
mtsolakiszen: This pattern is a bit confusing but what's happening is the errors object is getting instantiated and eventually returned with the `tap` call, then getting passed down into these methods which is pushing the errors onto it.
I don't love this pattern as it's a bit confusing however this seems to be being done in a few places in the code base.

lib/zendesk_apps_support/validations/requirements.rb
mtsolakiszen: Yep good idea, will update the code.

The edge case where `requirements['custom_objects'] = false` will skip over the validation for `custom_objects` which is fine since none are defined. The build will then fail and error in ZAM as a hash is expected to be returned. 
For this case we could handle the semantically incorrect values in ZAM or add validations to check that the requirements data structure contains the correct types. I'll add a comment to the custom objects implementation PR and we should follow this up in another story if we think its necessary

spec/validations/requirements_spec.rb
mtsolakiszen: I would say no for two reasons

The first is that a `ValidationError` being initialised doesn't necessarily mean it was added to the errors. Imagine a case where there has been a refactor to make this more functional and have the `ValidationError` object be returned from a method, but the definition of the parent method calling that forgets to push it onto the errors object. The test would pass, even though the validation was not performed correctly.

The second is that this would couple the test more closely with the implementation. Currently, the test has no knowledge of the `ValidationError` class, it only focuses that the output has the correct data in the correct structure. Should this class be renamed, moved, replaced or deleted, this test wouldn't care. As long as the output is what was expected. ie an array with the first object responding to `key` with the expected key and `data` with the expected hash

lib/zendesk_apps_support/validations/requirements.rb
keitheous: I understand 🤢And I also get what you mean about the existing pattern.

This is a **private method that accepts errors as an argument** with default value `[]`.

Does it mean that we do not need the third argument declared? 

Does it mean that we have `@errors` all along?

lib/zendesk_apps_support/validations/requirements.rb
keitheous: Awesome. Thank you.

spec/validations/requirements_spec.rb
keitheous: I feel you. 👐 

>Should this class be renamed, moved, replaced or deleted, this test wouldn't care. As long as the output is what was expected. ie an array with the first object responding to key with the expected key and data with the expected hash

This is like duck typing. as long as it quacks it is a duck. As long as it responds to `.first`, `.first.key` and `first.data`, it is an array containing an object with those methods.

I also understand your view on not coupling the test closely. When it is coupled and the class knows about ValidationError, it makes future change and improvements challenging. With this being a unit test, I always forget to test in isolation. I will be more deliberate moving forward. 

Thank you for the insight 👍

keitheous:
I saw your commit message -> "update method **gaurd** for consistency"

lib/zendesk_apps_support/validations/requirements.rb
mtsolakiszen: Not quite, its not the `@erros` that's being passed, its the `errors` var initialised in the `invalid_custom_objects` method during `[].tap | errors |`
Let's 🍖 about this but yeah, probably a pattern we should like into changing sooner rather than later

lib/zendesk_apps_support/validations/requirements.rb
```diff
@@ -92,23 +107,22 @@ def invalid_channel_integrations(requirements)
           end
         end
 
-        def invalid_custom_resources_schema(requirements)
-          custom_resources_schema = requirements['custom_resources_schema']
-          return unless custom_resources_schema
-          valid_schema_keys = %w[resource_types relationship_types]
+        def invalid_custom_objects(requirements)
+          custom_objects = requirements['custom_objects']
+          return unless custom_objects
```
seangoedecke: TODO: error if `requirements['custom_objects']` is `false` rather than just returning

PR #257, opened 2019-10-25
Add strings that refer to custom objects rather than resources

seangoedecke:
Just a rename to match the new term for custom objects, no other changes.


seangoedecke:
cc @zendesk/dingo @zendesk/localization

PR #262, opened 2019-12-02
[APPS-4892]Add translations

hansol-kim-zendesk:
@zendesk/wombat @zendesk/localization 

Added translations for when there are more than one oauth type parameters

### DIFF
--- a/config/locales/translations/zendesk_apps_support.yml
+++ b/config/locales/translations/zendesk_apps_support.yml
@@ -183,6 +183,11 @@ parts:
       title: "App builder job: missing oauth fields error"
       value: "Missing required oauth fields in manifest: %{missing_keys}"
   - translation:
+      key: "txt.apps.admin.error.app_build.too_many_oauth_parameters"
+      title: "App builder job: too_many_oauth_parameters"
+      value: "Too many parameters with type 'oauth': one permitted"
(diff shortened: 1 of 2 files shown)


config/locales/translations/zendesk_apps_support.yml
```diff
@@ -182,6 +182,11 @@ parts:
       key: "txt.apps.admin.error.app_build.oauth_keys.missing.other"
       title: "App builder job: missing oauth fields error"
       value: "Missing required oauth fields in manifest: %{missing_keys}"
+  - translation:
+      key: "txt.apps.admin.error.app_build.too_many_oauth_parameters"
+      title: "App builder job: too_many_oauth_parameters"
```
ovillegaszen: Can you please specify 'oauth' as an authorisation standard and is not to be translated.
Suggestion: "App builder job: upload error when there are more than one 'oauth' parameters. 'oauth' is an authorisation standard and is not to be translated"

hansol-kim-zendesk: @ovillegaszen Hi Oscar! I have changed to your suggestion :)

hansol-kim-zendesk: @zendesk/localization Hi Localization Team! Could someone have a look at this for us please?

PR #263, opened 2019-12-02
[DINGO-651] Add custom_object_requirements helper method

JeanMarcGoepfert:
After some discussion with @seangoedecke [here](https://github.com/zendesk/zendesk_app_market/pull/3854#discussion_r352382213) and offline, we decided the best place to put a method to check for custom object requirements is in zas. This adds it, and replaces some hard coded strings with consts.

### DIFF
--- a/lib/zendesk_apps_support/package.rb
+++ b/lib/zendesk_apps_support/package.rb
@@ -12,8 +12,8 @@ module ZendeskAppsSupport
     MANIFEST_FILENAME = 'manifest.json'
     REQUIREMENTS_FILENAME = 'requirements.json'
     CUSTOM_OBJECTS_REQUIREMENT_KEY = 'custom_objects'
-    CUSTOM_OBJECTS_REQUIREMENT_TYPE = 'custom_object_types'
-    CUSTOM_OBJECTS_RELATIONSHIP_TYPE = 'custom_object_relationship_types'
+    CUSTOM_OBJECTS_REQUIREMENT_TYPE_KEY = 'custom_object_types'
+    CUSTOM_OBJECTS_RELATIONSHIP_TYPE_KEY = 'custom_object_relationship_types'
 
     DEFAULT_LAYOUT = Erubis::Eruby.new(File.read(File.expand_path('../assets/default_template.html.erb', __FILE__)))
     DEFAULT_SCSS   = File.read(File.expand_path('../assets/default_styles.scss', __FILE__))
@@ -217,12 +217,10 @@ module ZendeskAppsSupport
 
     def self.has_custom_object_requirements?(requirements_hash)
       custom_objects = requirements_hash && requirements_hash[CUSTOM_OBJECTS_REQUIREMENT_KEY] || {}
-      types = custom_objects[CUSTOM_OBJECTS_REQUIREMENT_TYPE]
-      relationships = custom_objects[CUSTOM_OBJECTS_RELATIONSHIP_TYPE]
+      types = custom_objects[CUSTOM_OBJECTS_REQUIREMENT_TYPE_KEY]
+      relationships = custom_objects[CUSTOM_OBJECTS_RELATIONSHIP_TYPE_KEY]
 
-      return true if types && types.any?
-      return true if relationships && relationships.any?
-      false
+      (types && types.any?) || (relationships && relationships.any?)
     end
 
     def app_css
--- a/lib/zendesk_apps_support/package.rb
+++ b/lib/zendesk_apps_support/package.rb
@@ -11,9 +11,6 @@ module ZendeskAppsSupport
 
     MANIFEST_FILENAME = 'manifest.json'
     REQUIREMENTS_FILENAME = 'requirements.json'
-    CUSTOM_OBJECTS_REQUIREMENT_KEY = 'custom_objects'
-    CUSTOM_OBJECTS_REQUIREMENT_TYPE_KEY = 'custom_object_types'
-    CUSTOM_OBJECTS_RELATIONSHIP_TYPE_KEY = 'custom_object_relationship_types'
 
     DEFAULT_LAYOUT = Erubis::Eruby.new(File.read(File.expand_path('../assets/default_template.html.erb', __FILE__)))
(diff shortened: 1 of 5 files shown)


lib/zendesk_apps_support/package.rb
```diff
@@ -212,6 +215,13 @@ def has_requirements?
       has_file?(REQUIREMENTS_FILENAME)
     end
 
+    def self.has_custom_object_requirements?(requirements_hash)
+      custom_objects = requirements_hash&.dig(CUSTOM_OBJECTS_REQUIREMENT_KEY) || {}
+      return true if custom_objects[CUSTOM_OBJECTS_REQUIREMENT_TYPE]&.any?
+      return true if custom_objects[CUSTOM_OBJECTS_RELATIONSHIP_TYPE]&.any?
+      false
```
JeanMarcGoepfert: This isn't great. I want to return false when requirements_hash is any of these values: `nil`, `{}`, `{ custom_object_types: {}, custom_object_relationship_types: {} }`. We're using `present?` in the previous version which isn't available here. Keen to hear if anyone has a nicer way to do this.

lib/zendesk_apps_support/package.rb
```diff
@@ -212,6 +215,16 @@ def has_requirements?
       has_file?(REQUIREMENTS_FILENAME)
     end
 
+    def self.has_custom_object_requirements?(requirements_hash)
```
JeanMarcGoepfert: This isn't great. I want to return false when requirements_hash is any of these values: `nil`, `{}`, `{ custom_object_types: {}, custom_object_relationship_types: {} }`. We're using `present?` in the previous version which isn't available here. Keen to hear if anyone has a nicer way to do this.

lib/zendesk_apps_support/package.rb
```diff
@@ -11,6 +11,9 @@ class Package
 
     MANIFEST_FILENAME = 'manifest.json'
     REQUIREMENTS_FILENAME = 'requirements.json'
+    CUSTOM_OBJECTS_REQUIREMENT_KEY = 'custom_objects'
+    CUSTOM_OBJECTS_REQUIREMENT_TYPE = 'custom_object_types'
+    CUSTOM_OBJECTS_RELATIONSHIP_TYPE = 'custom_object_relationship_types'
```
fvzzy: nit: suffix these with `KEY` for consistency?

lib/zendesk_apps_support/package.rb
```diff
@@ -212,6 +215,16 @@ def has_requirements?
...
     end
 
+    def self.has_custom_object_requirements?(requirements_hash)
+      custom_objects = requirements_hash && requirements_hash[CUSTOM_OBJECTS_REQUIREMENT_KEY] || {}
+      types = custom_objects[CUSTOM_OBJECTS_REQUIREMENT_TYPE]
+      relationships = custom_objects[CUSTOM_OBJECTS_RELATIONSHIP_TYPE]
+
+      return true if types && types.any?
+      return true if relationships && relationships.any?
+      false
```
fvzzy: nit: `(types && types.any?) || (relationships && relationships.any?)`

spec/package_spec.rb
```diff
@@ -13,6 +13,74 @@
...
+    it 'should return false when custom object relationships are not present' do
+      requirements_hash = {
+        'custom_objects' => {
+          'custom_object_relationship_types' => []
+        }
+      }
+
+      expect(described_class.has_custom_object_requirements?(requirements_hash)).to be(false)
+    end
+
```
fvzzy: Edge case perhaps worth testing against: `should return true when custom object types are present but relationships types are not defined`, e.g.
```
      requirements_hash = {
        'custom_objects' => {
          'custom_object_types' => [
            {
              'key' => 'product',
              'schema' => { 'properties' => { 'id' => { 'type' => 'string' } } }
            }
          ],
		  'custom_object_relationship_types' => []
        }
```

Also the opposite, i.e. `should return true when custom object relationship types are present but custom object types are not defined`

lib/zendesk_apps_support/package.rb
JeanMarcGoepfert: Yeah, that's nicer. Thanks @fvzzy.

JeanMarcGoepfert: Actually, this won't work. If types, or relationships are nil, this will return nil. The linter doesn't allow double negation to coerce it to a bool either. I could do a `nil?` with !, but that doesn't seem much nicer than what was previously there.

fvzzy: `!!(nit: (types && types.any?) || (relationships && relationships.any?))`

Jk pls don't do this

spec/package_spec.rb
JeanMarcGoepfert: Done 👍

lib/zendesk_apps_support/package.rb
fvzzy: TIL https://stackoverflow.com/questions/12701869/nil-false-and-false-nil-in-ruby

JeanMarcGoepfert: Linter doesn't allow the `!!` anyway, probably for the best 😆

lib/zendesk_apps_support/package.rb
```diff
@@ -212,6 +215,13 @@ def has_requirements?
       has_file?(REQUIREMENTS_FILENAME)
     end
 
+    def self.has_custom_object_requirements?(requirements_hash)
+      return false if requirements_hash.nil?
+      types = requirements_hash.dig(CUSTOM_OBJECTS_REQUIREMENT_KEY, CUSTOM_OBJECTS_REQUIREMENT_TYPE_KEY) || []
+      relationships = requirements_hash.dig(CUSTOM_OBJECTS_REQUIREMENT_KEY, CUSTOM_OBJECTS_RELATIONSHIP_TYPE_KEY) || []
+      types.any? || relationships.any?
```
keitheous: This already looks a whole lot better than `a && b || c`. Well done. 

Wanted to point out that line 222 can be refactored to `(types | relationships).any?`
One can argue that this reduces readability 

Reference. http://www.chrisrolle.com/en/blog/array-coherences

PR #267, opened 2020-01-13
[DINGO-722] Remove IP checks

JeanMarcGoepfert:
IP checks are causing false positives for some users, so we've decided to remove them from ZAS.

Also installs required version of bundler in travis before_install hook to resolve failing builds.


.travis.yml
```diff
@@ -15,6 +15,8 @@ matrix:
   include:
     - rvm: 2.4.0
       env: COMMAND=rubocop
+before_install:
+  - gem install bundler:1.17.3
```
fvzzy: Perhaps grab the bundler version from the `Gemfile.lock`, e.g.  `gem install bundler:"$(grep -A 1 "BUNDLED WITH" Gemfile.lock | tail -n 1)"`

.travis.yml
```diff
@@ -15,6 +15,8 @@ matrix:
   include:
     - rvm: 2.4.0
       env: COMMAND=rubocop
+before_install:
```
mtsolakiszen: Why was this not needed before but needed now?

.travis.yml
JeanMarcGoepfert: Removed this, was sorted in https://github.com/zendesk/zendesk_apps_support/pull/269

.travis.yml
JeanMarcGoepfert: A change in travis I guess. I've removed this as it's sorted in https://github.com/zendesk/zendesk_apps_support/pull/269

token-cjg:
@JeanMarcGoepfert what are the risks of this change?

PR #268, opened 2020-01-16
[APPS-4953] Make parameter with kind oauth required when oauth section is present

fbvilela:
cc @zendesk/wattle 

### Description

Makes the oauth param required when the oauth section is present.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -62,7 +62,6 @@ en:
               other: Unsupported MIME types detected in %{file_names}.
             multiple_channel_integrations: Specifying multiple channel integrations
               in requirements.json is not supported.
-            too_many_oauth_parameters: "Too many parameters with type 'oauth': one permitted"
             oauth_parameter_required: "Please upgrade to our new oauth format. Learn more: %{link}"
             invalid_cr_schema_keys:
               one: 'Custom resources schema contains an invalid key: %{invalid_keys}'
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -62,6 +62,8 @@ en:
               other: Unsupported MIME types detected in %{file_names}.
             multiple_channel_integrations: Specifying multiple channel integrations
               in requirements.json is not supported.
+            too_many_oauth_parameters: "Too many parameters with type 'oauth': one permitted"
+            oauth_parameter_required: "Please upgrade to our new oauth format. Learn more: %{link}"
             invalid_cr_schema_keys:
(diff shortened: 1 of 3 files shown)


config/locales/en.yml
```diff
@@ -63,6 +63,7 @@ en:
             multiple_channel_integrations: Specifying multiple channel integrations
               in requirements.json is not supported.
             too_many_oauth_parameters: "Too many parameters with type 'oauth': one permitted"
+            oauth_parameter_required: "At least one parameter with the type 'oauth' is required"
```
jalee-zendesk: Only one `oauth` parameter is currently allowed. "At least one parameter" might imply that it is possible to have more than one perhaps?

hansol-kim-zendesk:
Code looks good to me, as long as the travis is passing!

fbvilela:
cc @zendesk/localization

hansol-kim-zendesk:
Once linting is fixed, they are good to go!

config/locales/en.yml
```diff
@@ -62,6 +62,8 @@ en:
               other: Unsupported MIME types detected in %{file_names}.
             multiple_channel_integrations: Specifying multiple channel integrations
               in requirements.json is not supported.
+            too_many_oauth_parameters: "Too many parameters with type 'oauth': one permitted"
```
lcampbell-zendesk, line 65: This doesn't appear to be referenced below.

fbvilela: Logan, yes... this was a rebase issue. I removed this line 👍

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -197,14 +197,21 @@ def private_marketing_app_error(manifest)
...
 
           if missing.any?
-            ValidationError.new('oauth_keys.missing', missing_keys: missing.join(', '), count: missing.length)
+            oauth_errors << \
+              ValidationError.new('oauth_keys.missing', missing_keys: missing.join(', '), count: missing.length)
+          end
+
+          unless manifest.parameters.any? { |param| param.type == 'oauth' }
+            oauth_errors << ValidationError.new('oauth_parameter_required',
+                                                link: 'https://developer.zendesk.com/apps/docs/developer-guide/manifest#oauth')
```
waferbaby, line 212: Idea: Should we move the link to a constant, in case we need to change it?

fbvilela: done

PR #272, opened 2020-02-25
[POC] Support to bundle_oauth and bundle_integration_key

7ail:
### Description

Install App

In order to utilise the external OAuth Client created here, changes need to be made in ZAM.

The flow we are striving to achieve is as follows
image

### CHANGES TO ZAS

Since we are dependent on an OAuthClient generated outside of ZAM, changes need to be made to the app manifest.json file to enable the app to make use of it. The plan is to create two new custom hidden fields like the following 

```
"parameters": [
  {
    "name": "bundle_token",
    "type": "bundle_oauth"
  },
  {
    "name": "bundle_integration_key",
    "type": "bundle_integration_key"
  }
]
```

###Features

* Both must exist together
* Both fields are hidden
* `bundle_oauth`
** This field will store the connection_uuid like the pre-existing `oauth` type
* `bundle_integration_key`
** This field will store the integration key which is specified during the ZIS-Bundle upload (removing the need to rely on procedurally generated versions from ZAM)

Currently, if we rely on the OAuthClient generated in ZAM, the flow sequence looks something like this:
image

With these changes, ZAP will need some underlying logic to recognise the existence of both of these parameter types and utilise their values to access ZIS-Connections for the `access_token`. This eliminates the need to query ZAM for the `integration_key`
image

DONE WHEN

- [x] ZAS is able to validate both `bundle_oauth` and `bundle_integration_key`
- [x] Translations are ignored in this POC

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -63,6 +63,10 @@ en:
             multiple_channel_integrations: Specifying multiple channel integrations
               in requirements.json is not supported.
             too_many_oauth_parameters: "Too many parameters with type 'oauth': one permitted"
+            too_many_bundle_oauth: "Too many parameters with type 'bundle_oauth': one permitted"
+            too_many_bundle_integration_key: "Too many parameters with type 'bundle_integration_key': one permitted"
+            missing_bundle_oauth: "Parameter of type 'bundle_integration_key' exist without 'bundle_oauth'"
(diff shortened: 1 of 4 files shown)


yoshdog:
It should be noted that although we are proposing this for our new Shopify integration on ZIS, it will ideally be the approach we would use for integrations built on ZIS that require an OAuth + Sidebar app.

PR #274, opened 2020-04-30
Make `location_icons` a public method

waferbaby:
We need to be able to call the `location_icons` method from within ZAM, and don't want to rely on it always working via a semi-hacky `send` call.


token-cjg:
@waferbaby , what is the risk of this change?

waferbaby:
@token-cjg Low - updated, sorry that got dropped from the PR!

PR #275, opened 2020-04-30
Bump to v4.29.1

waferbaby:
Makes `location_icons` a public method.


lcampbell-zendesk:
If this is branched off master does it include the make-public change?

waferbaby:
@lcampbell-zendesk That was already merged to master.

PR #276, opened 2020-05-13
Created a new location for visits on web

mikhail-ev:


szymzet:
cc @pgrzesiak

token-cjg:
@mikhail-ev , what is the risk of this change?

PR #280, opened 2020-07-02
[APPS-5183]Add translation

hansol-kim-zendesk:
/cc @zendesk/wattle @zendesk/dingo @zendesk/localization 

### Description
Add translation for when the parameter type is oauth and it is set to secure. 

### References
Link to a JIRA or GitHub issue here if relevant

#### Before merging this PR

### Risks
* [RUNTIME] No
* [LOW] Adding translation


hansol-kim-zendesk:
Thank you @hanli625 😁

PR #281, opened 2020-07-02
[APPS-5183]Add validation

hansol-kim-zendesk:
💐

/cc @zendesk/wattle

### Description
Add validation for manifest for parameter type oauth. This cannot be seucre.

### References
https://zendesk.atlassian.net/browse/APPS-5183

#### Before merging this PR

### Risks
* [RUNTIME] No
* [Low] Adding validation

### DIFF
--- a/lib/zendesk_apps_support/validations/manifest.rb
+++ b/lib/zendesk_apps_support/validations/manifest.rb
@@ -39,6 +39,33 @@ module ZendeskAppsSupport
           errors << oauth_error(manifest)
           errors << default_locale_error(manifest, package)
 
+          errors << validate_parameters(manifest)
+
+          if manifest.requirements_only? || manifest.marketing_only?
+            errors << ban_location(manifest)
+            errors << ban_framework_version(manifest)
+          else
+            errors << validate_location(package)
+            errors << missing_framework_version(manifest)
+            errors << invalid_version_error(manifest)
+          end
+
+          errors << ban_no_template(manifest) if manifest.iframe_only?
+
+          errors.flatten.compact
+        end
+
+        def validate_location(package)
+          manifest = package.manifest
+          errors = []
+          errors << missing_location_error(package)
+          errors << invalid_location_error(package)
+          errors << invalid_v1_location(package)
+          errors << location_framework_mismatch(manifest)
+        end
+
+        def validate_parameters(manifest)
+          errors = []
           if manifest.marketing_only?
             errors.push(*marketing_only_errors(manifest))
           else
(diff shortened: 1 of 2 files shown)


lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -66,6 +55,37 @@ def collate_manifest_errors(package)
           errors.flatten.compact
         end
 
+        def validate_location(package)
+          manifest = package.manifest
+          errors = []
+          errors << missing_location_error(package)
+          errors << invalid_location_error(package)
+          errors << invalid_v1_location(package)
+          errors << location_framework_mismatch(manifest)
```
lcampbell-zendesk, line 64: I would like to see `errors` explicitly returned at the end of this method. It might be returned by `<<` but if someone modified this method and added some additional logic they might be surprised when their code stops working. Having `return errors` (or rubocop will probably make you write it as just `errors`) will make the method a little more resilient to change.

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -39,25 +39,14 @@ def collate_manifest_errors(package)
...
+          errors << validate_parameters(manifest)
 
           if manifest.requirements_only? || manifest.marketing_only?
             errors << ban_location(manifest)
             errors << ban_framework_version(manifest)
           else
-            errors << missing_location_error(package)
-            errors << invalid_location_error(package)
-            errors << invalid_v1_location(package)
+            errors << validate_location(package)
```
lcampbell-zendesk, line 48: Because `validate_location` is returning an array I'm surprised that you don't need to do `.concat` here... I see down below that we're doing `flatten.compact` on the whole thing.. That actually gives us a lot of room to improve this whole thing. We could remove every instance of `<<` here by instead making the whole method an array literal with method calls inside it. Like:

``` ruby
[
  oauth_error(manifest),
  default_locale_error(manifest, package),
  validate_parameters(manifest),
  if manifest.requirements_only? || manifest.marketing_only?
    [
      ban_location(manifest),
      ban_framework_version(manifest)
    ]
  else
    [ ... ]
  end,
  ...
]
```

Because `if` is an expression in ruby it should work nicely, and remove a lot of this duplication. I should've seen this earlier, but I had assumed there was some structure to the errors that needed to be maintained.

Up to you if you'd like to try this style of change.

lib/zendesk_apps_support/validations/manifest.rb
hansol-kim-zendesk: done!

lib/zendesk_apps_support/validations/manifest.rb
hansol-kim-zendesk: We did this together and use the same approach to refactor other methods

hansol-kim-zendesk:
Tests are failing because of the translation

PR #296, opened 2021-02-15
[DINGO-1233] Add SonarQube config

fvzzy:
🐕 

@zendesk/dingo

### Description
Adds SonarQube. Most of this is just copy-pasted from https://github.com/zendesk/marketplace_payment_service/pull/139, with some config tweaks from the [Onboarding Steps](https://zendesk.atlassian.net/wiki/spaces/BRE/pages/1326025188/Analysing+your+Repository+with+SonarQube). 

~Dashboard at: https://sonarqube.internaltools-staging-use1.zende.sk/dashboard?id=zendesk_apps_support.~

### References
* JIRA: https://zendesk.atlassian.net/browse/DINGO-1233

### Risks
* Low - CI starts failing.

### DIFF
--- a/.github/workflows/sonarqube.yml
+++ b/.github/workflows/sonarqube.yml
@@ -12,6 +12,8 @@ jobs:
     steps:
     - name: SonarQube scan Master
       uses: zendesk/checkout@v2
+      with:
+        fetch-depth: 0
       if: github.event_name == 'push' && github.ref == 'refs/heads/master'
       run: |
         sonar-scanner \
--- a/.github/workflows/sonarqube.yml
+++ b/.github/workflows/sonarqube.yml
@@ -10,8 +10,6 @@ jobs:
   main:
     runs-on: ubuntu-latest
     steps:
-    - name: Setup SonarQube
-      uses: warchant/setup-sonar-scanner@v3
     - name: SonarQube scan Master
       uses: zendesk/checkout@v2
       if: github.event_name == 'push' && github.ref == 'refs/heads/master'
--- a/.github/workflows/sonarqube.yml
+++ b/.github/workflows/sonarqube.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   main:
-    runs-on: zendesk-stable
+    runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/sonarqube.yml
```diff
@@ -0,0 +1,38 @@
...
+
+on:
+  push:
+    branches: master
+  pull_request:
+    branches: master
+
+jobs:
+  main:
+    runs-on: zendesk-stable
```
token-cjg, line 11: I think here you'd be after `runs-on: [self-hosted, zendesk-stable]`.

fvzzy: There might be a couple of different syntaxes that work here, I've used this format with success before on [Apps Asset Service](https://github.com/zendesk/apps_asset_service/pull/17/files#diff-a5e11e7bedd328777f4897395a711a4492d413b70b12479230470ec608d96792R11) and [Developer Docs](https://github.com/zendesk/developer-docs/pull/141/files#diff-a5e11e7bedd328777f4897395a711a4492d413b70b12479230470ec608d96792R11)

fvzzy: Actually sorry I think you're right Chris. Should I update those two repos to be `[self-hosted, zendesk-stable]` as well?

fvzzy: Turns out as a public repo this needs to use a Github-hosted runner rather than one of our self-hosted ones

.github/workflows/sonarqube.yml
```diff
@@ -8,10 +8,12 @@ on:
 
 jobs:
   main:
-    runs-on: ubuntu-latest
+    runs-on: zendesk-stable
```
AeroCross, line 11: @hansol-kim-zendesk, I _think_ the diff we got was borked, because [this shouldn't work](https://zendesk.slack.com/archives/CV1GMGC0H/p1613352547158400).

Indeed, [what was configured before is the recommended setting](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources).

PR #308, opened 2021-08-11
[VEG-599] Support flexible parameter in manifest validation

token-cjg:
🐕

/cc @zendesk/dingo @zendesk/vegemite 

### Description

As part of the responsive sidebar apps work, we'd like our manifests to validate an additional "flexible" parameter that will sit within the location options.

This will be a latent option for all locations across all hosts, but in practice will only be supported for the Support ticket_sidebar and new_ticket_sidebar locations.

### References
https://zendesk.atlassian.net/browse/VEG-599

#### Before merging this PR
- [x] Add more unit tests

### Risks
* [RUNTIME] Can this change affect apps rendering for a user? Technically, no. But this will be used as the foundation for future changes in Support, that will use this new attribute. Those particular changes definitely will affect apps rendering for a user, as that is the key idea of responsive sidebar locations.
* Low.  Adds an additional validation for a new parameter.  Change should be backwards compatible.


sarmadsangi (approved):
This works for ZAT as expected it seems. We will need @zendesk/dingo review on this to see if this change is sufficient for ZAM to emit out that we require and also if validations work on that end as expected.

fbvilela (approved):

PR #312, opened 2021-08-26
v4.29.10

fbvilela:
🐕

/cc @zendesk/dingo

### Description
Version bump + dropping support for Ruby 2.4.


token-cjg (approved):
This should be okay, as:

* as of MacOS Catalina and Big Sur, default system ruby is 2.6.3,
* and RubyInstaller https://rubyinstaller.org/ which seems to be the Windows Way ™️ to install ruby currently gives default ruby versions of 2.7.x or 3.0.x

fvzzy (approved):

PR #326, opened 2022-05-16
[DINGO-1723] Add url validation

chris55miao:
🐕

/cc @zendesk/dingo

### Description
A private app (and potentially any marketplace app) can contain JavaScript in the manifest.json file which gets executed from the {subdomain}.zendesk.com domain.

STEPS TO REPRODUCE
1. Update the url field in an app in the manifest.json file to "url": "javascript:alert(document.domain)"
2. Package the app and Upload the private App to your account.
3. One the app is installed, click on the App and as the url field is populated, you will see an option to click on the "developer website" link.
4. Click on the link and the javascript will trigger the alert within the context of the accounts subdomain.

An attacker can use this vulnerability to construct a request that, if issued by another application user, will cause JavaScript code supplied by the attacker to execute within the user's browser in the context of that user's session with the application, steal user's cookies, etc.

### References
https://zendesk.atlassian.net/browse/DINGO-1723

### Risks
* [low] validate url

### DIFF
--- a/config/locales/translations/zendesk_apps_support.yml
+++ b/config/locales/translations/zendesk_apps_support.yml
@@ -412,3 +412,7 @@ parts:
       key: "txt.apps.admin.error.app_build.oauth_parameter_cannot_be_secure"
       title: "oauth parameter cannot be set to be secure."
       value: "oauth parameter cannot be set to be secure."
+  - translation:
+      key: "txt.apps.admin.error.app_build.invalid_url"
+      title: "App builder job: this value needs to be a valid URL, but something else was passed in. placeholder value is taken from user input. You can translate as: The value %{field} must be a valid URL... to avoid any gender issues."
+      value: "%{field} must be a valid URL, got \"%{value}\"."
--- a/lib/zendesk_apps_support/validations/manifest.rb
+++ b/lib/zendesk_apps_support/validations/manifest.rb
@@ -39,6 +39,7 @@ module ZendeskAppsSupport
             type_checks(manifest),
             oauth_error(manifest),
             default_locale_error(manifest, package),
+            validate_urls(manifest),
             validate_parameters(manifest),
             if manifest.requirements_only? || manifest.marketing_only?
               [ ban_location(manifest),
@@ -61,6 +62,18 @@ module ZendeskAppsSupport
             invalid_v1_location(package),
             location_framework_mismatch(manifest)
           ]
+        end
+
+        def validate_urls(manifest)
+          errors = []
+          if manifest.terms_conditions_url
+            errors << validate_url(manifest.terms_conditions_url, "terms_conditions_url")
+          end
+
+          if manifest.author
+            errors << validate_url(manifest.author["url"], "author url")
+          end
+          errors
         end
 
         def validate_parameters(manifest)
@@ -414,6 +427,12 @@ module ZendeskAppsSupport
           return if [false, true].include? no_template
           unless no_template.is_a?(Array) && manifest.no_template_locations.all? { |loc| Location.find_by(name: loc) }
             ValidationError.new(:invalid_no_template)
+          end
+        end
+
+        def validate_url(value, label_for_error)
+          unless value.nil? || value.match(/^https?:\/\//)
+            ValidationError.new(:invalid_url, field: label_for_error, value: value)
           end
         end
       end
--- a/lib/zendesk_apps_support/validations/manifest.rb
+++ b/lib/zendesk_apps_support/validations/manifest.rb
@@ -431,9 +431,14 @@ module ZendeskAppsSupport
         end
 
         def validate_url(value, label_for_error)
-          unless value.nil? || value.match(/^https?:\/\//)
+          return if value.nil?
+
+          uri = URI.parse(value)
+          unless uri.is_a?(URI::HTTP) && !uri.host.nil?
             ValidationError.new(:invalid_url, field: label_for_error, value: value)
           end
+        rescue URI::InvalidURIError
+          ValidationError.new(:invalid_url, field: label_for_error, value: value)
         end
       end
     end
--- a/lib/zendesk_apps_support/validations/manifest.rb
+++ b/lib/zendesk_apps_support/validations/manifest.rb
@@ -66,12 +66,14 @@ module ZendeskAppsSupport
 
         def validate_urls(manifest)
           errors = []
-          if manifest.terms_conditions_url
-            errors << validate_url(manifest.terms_conditions_url, "terms_conditions_url")
+
+          if manifest.terms_conditions_url && !valid_url?(manifest.terms_conditions_url)
+            errors << ValidationError.new(:invalid_url, field: 'terms_conditions_url', value: manifest.terms_conditions_url)
           end
 
-          if manifest.author
-            errors << validate_url(manifest.author["url"], "author url")
+          author = manifest.author
+          if author && author["url"] && !valid_url?(author["url"])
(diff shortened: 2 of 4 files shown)


config/locales/translations/zendesk_apps_support.yml
```diff
@@ -412,3 +412,7 @@ parts:
       key: "txt.apps.admin.error.app_build.oauth_parameter_cannot_be_secure"
       title: "oauth parameter cannot be set to be secure."
       value: "oauth parameter cannot be set to be secure."
+  - translation:
+      key: "txt.apps.admin.error.app_build.invalid_url"
+      title: "App builder job: this value needs to be a valid URL, but something else was passed in. placeholder value is taken from user input. You can translate as: The value %{field} must be a valid URL... to avoid any gender issues."
```
JeanMarcGoepfert, line 417: This title doesn't make much sense to me 😅, generally they should just be a short description of what the error means. In this case  something like 'The url provided was not valid' should do.

Frederick888 (approved):
Wow this is a great catch. Thank you!

config/locales/translations/zendesk_apps_support.yml
chris55miao: I just copied the invalid_Sstring translation 😅. Happy to change it

JeanMarcGoepfert: I see. Happy to leave it as is since there's a few with a similar description already. Just seems pretty weird to me that we mention the app builder job here - that's in a completely different code base 🤷

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -416,6 +429,12 @@ def no_template_format_error(manifest)
             ValidationError.new(:invalid_no_template)
           end
         end
+
+        def validate_url(value, label_for_error)
+          unless value.nil? || value.match(/^https?:\/\//)
```
JeanMarcGoepfert, line 434: Have you confirmed with product that we require https urls now? There's a lot of stuff like `http://foo.com`, `www.foo.com` and just `foo.com`s in the db at the moment.

config/locales/translations/zendesk_apps_support.yml
chris55miao: maybe some historical issue

lib/zendesk_apps_support/validations/manifest.rb
chris55miao: interesting, i check that if the url is www.xxxx.com, the redirection will treat that as a path of the url
in my case
```
"url": "www.urlencoder.org"
```
redirect me to: `https://z3nchristinawu.zendesk-staging.com/support/apps/www.urlencoder.org`

image

that mean we dont support that syntax very well?

chris55miao: currently `/^https?:\/\//` match both `http://` and `https:`

I wonder if we need to restrict them to `https` only

chris55miao:
@zendesk/localization

lib/zendesk_apps_support/validations/manifest.rb
chris55miao: Thanks @mmassaki , yep, we found many http and https in the database, so we should keep both

chris55miao: > Use URI class to validate urls

Done

spec/validations/manifest_spec.rb
```diff
@@ -692,4 +692,55 @@ def create_package(parameter_hash)
...
+  context 'url is not HTTP or HTTPS' do
+    before do
+      @manifest_hash = {
+        'termsConditionsURL' => 'javascript:alert("terms_conditions_url")',
+        'author' => { 'url' => 'javascript:alert("author_url")' }
+      }
+    end
+
+    it 'terms_conditions_url should have an error' do
+      expect(@package).to have_error(/terms_conditions_url must be a valid URL/)
```
mmassaki, line 705: Should we use the complete error message instead of a regex so we assert that the value is being returned to the user? This applies to the tests bellow as well.

spec/validations/manifest_spec.rb
```diff
@@ -692,4 +692,55 @@ def create_package(parameter_hash)
...
+    end
+
+    it 'terms_conditions_url should not have an error' do
+      expect(@package).not_to have_error(/terms_conditions_url must be a valid URL/)
+    end
+
+    it 'author.url should not have an error' do
+      expect(@package).not_to have_error(/author url must be a valid URL/)
+    end
+  end
```
mmassaki, line 745: How about edge cases? Is it possible for a manifest to have `nil` or `''` as url values?

mmassaki (commented):
Looking good!

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -416,6 +429,17 @@ def no_template_format_error(manifest)
...
+        def validate_url(value, label_for_error)
+          return if value.nil?
+
+          uri = URI.parse(value)
+          unless uri.is_a?(URI::HTTP) && !uri.host.nil?
+            ValidationError.new(:invalid_url, field: label_for_error, value: value)
+          end
+        rescue URI::InvalidURIError
+          ValidationError.new(:invalid_url, field: label_for_error, value: value)
+        end
```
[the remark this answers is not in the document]
chris55miao: I was following the same format as validate_string() method in the class
happy to change to the way you suggested, it is clearer

thanks~

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -416,6 +429,17 @@ def no_template_format_error(manifest)
             ValidationError.new(:invalid_no_template)
           end
         end
+
+        def validate_url(value, label_for_error)
+          return if value.nil?
```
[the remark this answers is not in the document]
chris55miao: this could be possible because url in author object is optional

spec/validations/manifest_spec.rb
chris55miao: done

spec/validations/manifest_spec.rb
chris55miao: good point, done

spec/validations/manifest_spec.rb
```diff
@@ -692,4 +692,88 @@ def create_package(parameter_hash)
...
+  context 'url is not HTTP or HTTPS' do
+    before do
+      @manifest_hash = {
+        'termsConditionsURL' => 'javascript:alert("terms_conditions_url")',
+        'author' => { 'url' => 'javascript:alert("author_url")' }
+      }
+    end
+
+    it 'terms_conditions_url should have an error' do
+      expect(@package).to have_error("terms_conditions_url must be a valid URL, got \"javascript:alert(\"terms_conditions_url\")\".")
```
mmassaki, line 705: nit: we can avoid escaping `"` by using single quotes. e.g. `'terms_conditions_url must be a valid URL, got "javascript:alert("terms_conditions_url")".'`

mmassaki (approved):
Thank you for addressing my comments.

spec/validations/manifest_spec.rb
chris55miao: done

chris55miao: thanks so much for being looking into detail

hanli625 (approved):
👍 for the string. Thanks!

JeanMarcGoepfert (approved):

PR #340, opened 2022-08-15
[APPS-6388] Bump nokogiri to fix security vulnerability

aishkan:
💐

/cc @zendesk/wattle

### Description
Bump nokogiri to fix security vulnerability and drop support for ruby 2.5 since Nokogiri >=1.3.6 supports ruby >=2.6.

### References
https://zendesk.atlassian.net/browse/APPS-6388

#### Before merging this PR

### Risks
* [RUNTIME] Can this change affect apps rendering for a user? No
* [HIGH | medium | low] What features does this touch? Low. Only bumps Nokogiri gem.


chris55miao (approved):

.github/workflows/actions.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby-version: [2.6.8, 2.5.9]
+        ruby-version: [2.6.8]
```
Frederick888, line 14: Could you add `2.7.5` (ZAM) to this matrix? :)

mmassaki, line 14: At the moment ZAS specifies `s.required_ruby_version = '>= 2.0'`. Should we bump the ruby version requirement if we are not testing it against older ruby versions?

mmassaki: Thinking about that, should we replace `2.6.8` with `2.6.10` and add `3.0.4` and `3.1.2` as well?

mmassaki (approved):

mmassaki (approved):

PR #345, opened 2022-10-18
[APPS-6494] Remove deprecated framework versions

mmassaki:
💐

/cc @zendesk/wattle

### Description
Removes support for deprecated app framework versions.

### References
https://zendesk.atlassian.net/browse/APPS-6494

#### Before merging this PR

### Risks
* [RUNTIME] Can this change affect apps rendering for a user? No, apps with deprecated versions are not being rendered.
* [low] What features does this touch? App framework version check

### DIFF
--- a/spec/app_version_spec.rb
+++ b/spec/app_version_spec.rb
@@ -31,9 +31,13 @@ describe ZendeskAppsSupport::AppVersion do
   end
 
   describe 'the deprecated version' do
-    let(:sample_version) { ZendeskAppsSupport::AppVersion::DEPRECATED.sample }
+    before do
+      stub_const('ZendeskAppsSupport::AppVersion::DEPRECATED', ['1.0'])
+      stub_const('ZendeskAppsSupport::AppVersion::TO_BE_SERVED', ['1.0'])
+    end
+
     subject do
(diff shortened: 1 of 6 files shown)


chris55miao (approved):

spec/app_version_spec.rb
```diff
@@ -31,9 +31,13 @@
   end
 
   describe 'the deprecated version' do
-    let(:sample_version) { ZendeskAppsSupport::AppVersion::DEPRECATED.sample }
+    before do
+      stub_const('ZendeskAppsSupport::AppVersion::DEPRECATED', ['1.0'])
```
jalee-zendesk, line 35: Nit: Not sure if it matters for these tests, but should the deprecated version be different to the "to be served" version?

jalee-zendesk (approved):

spec/app_version_spec.rb
mmassaki: Not necessarily, the `TO_BE_SERVED` is a superset of the `DEPRECATED` version, so they can be the same as well.
As a good practice, tests should have only the minimal data to run, so it's clear and easy to understand.

PR #355, opened 2023-05-10
add ruby 3.1 and 3.2 support, drop ruby 2.6

razumau:
### Description
Adds Ruby 3.1 and 3.2 to the test matrix, drops Ruby 2.6, updates tests dependencies (to versions with support for recent Ruby versions).
This required a couple of changes:
— keyword arguments in i18n’s `translate` method (because of that, I’ve added a lower limit for i18n’s version in the gemspec—[this release](https://github.com/ruby-i18n/i18n/releases/tag/v1.7.1));
— Hash refinements are removed in recents i18n versions (https://github.com/ruby-i18n/i18n/pull/573), so instead we use `I18n::Utils.except` in tests now;
— In newer `URI` versions, host for "https://" is an empty string, not nil, so we check for both.

I’ve also cleaned up tests workflow and added a `tests_successful` wrapper.

#### Before merging this PR

### Risks
* [RUNTIME] Can this change affect apps rendering for a user? No direct changes to rendering.
* [HIGH | medium | low] What features does this touch? i18n dependency


lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -434,7 +434,8 @@ def no_template_format_error(manifest)
 
         def valid_url?(value)
           uri = URI.parse(value)
-          uri.is_a?(URI::HTTP) && !uri.host.nil?
+          host_empty = uri.host.nil? || uri.host == ''
+          uri.is_a?(URI::HTTP) && !host_empty
```
razumau, line 438: In newer URI versions, host for "https://" is an empty string, not `nil`.

lib/zendesk_apps_support/manifest/no_override_hash.rb
```diff
@@ -18,7 +18,7 @@ def message
 
           # if the error contains the word `_legacy` in the second sentence, let's
           # only use the first one.
-          if [original, attempted].any? { |val| val =~ /_legacy/ }
+          if [original, attempted].any? { |val| val.is_a?(String) && val =~ /_legacy/ }
```
razumau, line 21: `=~` for non-strings has been removed from recent Ruby versions (and was deprecated—and often meaningless—before), so now we compare only strings to this regex.

spec/package_spec.rb
```diff
@@ -300,7 +300,7 @@ def build_app_source_with_files(files)
 
         it 'includes zh-cn in translations' do
           expected_translations = JSON.parse(File.read('spec/translations/zh-cn.json'))
-          expect(package.send(:translations)['zh-cn'].except('custom1')).to eq(expected_translations)
+          expect(I18n::Utils.except(package.send(:translations)['zh-cn'], 'custom1')).to eq(expected_translations)
```
razumau, line 303: Hash refinements are removed in recents i18n versions (https://github.com/ruby-i18n/i18n/pull/573).

spec/package_spec.rb
```diff
@@ -319,7 +319,7 @@ def build_app_source_with_files(files)
 
         it 'removes zendesk-specific keys in translations' do
           expected_translations = JSON.parse(File.read('spec/translations/zh-cn.json'))
-          expect(package.send(:translations)['zh-cn'].except('custom1')).to eq(expected_translations)
+          expect(I18n::Utils.except(package.send(:translations)['zh-cn'], 'custom1')).to eq(expected_translations)
```
razumau, line 322: Hash refinements are removed in recents i18n versions (https://github.com/ruby-i18n/i18n/pull/573).

lib/zendesk_apps_support/manifest/no_override_hash.rb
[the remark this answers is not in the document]
razumau: No: `val` is sometimes (at least in tests) a Hash (of all things), so that also fails with `TypeError: no implicit conversion of Hash into String`.

Shengming-Yan (approved):

zendesk_apps_support.gemspec
```diff
@@ -2,18 +2,18 @@
...
+  s.version     = '4.39.0'
   s.license     = 'Apache License Version 2.0'
   s.authors     = ['James A. Rosen', 'Likun Liu', 'Sean Caffery', 'Daniel Ribeiro']
   s.email       = ['dev@zendesk.com']
   s.homepage    = 'http://github.com/zendesk/zendesk_apps_support'
   s.summary     = 'Support to help you develop Zendesk Apps.'
   s.description = s.summary
 
-  s.required_ruby_version = Gem::Requirement.new('>= 2.6', '< 3.1')
+  s.required_ruby_version = Gem::Requirement.new('>= 2.7')
```
mmassaki, line 13: I think we can drop v2.7 support given it's past EOL now.

mmassaki (approved):

PR #364, opened 2024-08-21
[APPS-6948] Add manifest validation to deprecate password type param

aishkan:
💐

/cc @zendesk/wattle

### Description
Add manifest validation to disallow password type params.
The check is based on enabling `skip_password_parameter_check` to ensure this validation works only for new app submission. The value will be passed in from Apps approval and ZAM.

The skip_password_parameter_check value will be passed in from Apps Approval gem [here](https://github.com/zendesk/apps_approval/blob/main/lib/apps_approval/app_validation.rb#L21).

### References
https://zendesk.atlassian.net/browse/APPS-6946

### Risks
* [RUNTIME] No
* [low] Only adds an extra validation in manifest that works dependant on skip_password_parameter_check. Manifest validation can fail for apps.


hansol-kim-zendesk (approved):

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -22,7 +22,7 @@ def call(package)
             return [ValidationError.new(:missing_manifest)]
           end
 
-          collate_manifest_errors(package)
+          collate_manifest_errors(package, opts.fetch(:skip_password_parameter_check, true) )
```
JeanMarcGoepfert, line 25: Maybe I've been a manager too long, but I'm finding this a bit hard to follow. We default the `skip_password_parameter` value to false [here](https://github.com/zendesk/zendesk_apps_support/pull/364/files#diff-3bda7d468632d3fefde0a7a1b79ff0e590a24cdeaa753db9fa28493d6f6c2a5eR34), then default it to true [here](https://github.com/zendesk/zendesk_apps_support/pull/364/files#diff-721c968da45bad1e120fc0e48af8c3a66d2815bf105dc3ca915c48d6f823ddceR25) then apply the rule conditionally with a double negative.

Would it be simpler to change the variable name to `apply_parameter_check` and just give it the appropriate default value in one place?

aishkan: yeah that would be more readable. I have updated it now.

JeanMarcGoepfert (approved):

lib/zendesk_apps_support/validations/manifest.rb
mmassaki: Could we get the PR description updated with the new implementation using `apply_parameter_check` please?

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -13,7 +13,7 @@ module Manifest
       OAUTH_MANIFEST_LINK = 'https://developer.zendesk.com/apps/docs/developer-guide/manifest#oauth'
 
       class << self
-        def call(package)
+        def call(package, opts={})
```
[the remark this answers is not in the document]
aishkan: Changed it now

lib/zendesk_apps_support/package.rb
```diff
@@ -31,9 +31,9 @@ def initialize(dir, is_cached = true)
       @warnings = []
     end
 
-    def validate(marketplace: true, skip_marketplace_translations: false)
+    def validate(marketplace: true, skip_marketplace_translations: false, apply_password_parameter_check: false)
```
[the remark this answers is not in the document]
aishkan: I have added the tests now

spec/validations/manifest_spec.rb
```diff
@@ -4,6 +4,7 @@
 require 'tmpdir'
 
 describe ZendeskAppsSupport::Validations::Manifest do
+
```
mmassaki, line 7: nit: it seems other specs don't have this additional line break. How about reverting this one and removing the the additional line break on the new test examples?

spec/validations/manifest_spec.rb
```diff
@@ -556,7 +557,7 @@ def create_package(parameter_hash)
             'access_token_uri' => '4'
           }
         }
-        expect(@package).to have_error(/Please upgrade to our new oauth format/)
+        expect(create_package(@manifest_hash)).to have_error(/Please upgrade to our new oauth format/)
```
mmassaki, line 560: How was this working before? 🤔

lib/zendesk_apps_support/validations/manifest.rb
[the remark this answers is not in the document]
mmassaki: Sorry, I think my comment wasn't clear. Could we have it as a keyword arg so it's clear when calling this method?

mmassaki:
This moved to https://github.com/zendesk/zendesk_apps_support/pull/366

PR #366, opened 2024-09-17
[APPS-6946] Add manifest validation to deprecate password type param

aishkan:
/cc @zendesk/wattle

Description
Add manifest validation to disallow password type params.
The check is based on enabling _password_parameter_check to ensure this validation works only for new app submission. The value will be passed in from Apps approval and ZAM.

The apply _password_parameter_check value will be passed in from Apps Approval gem [here](https://github.com/zendesk/apps_approval/blob/main/lib/apps_approval/app_validation.rb#L21).

References
https://zendesk.atlassian.net/browse/APPS-6946

Risks
[RUNTIME] No
[low] Only adds an extra validation in manifest that works dependant on skip_password_parameter_check. Manifest validation can fail for apps.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -149,6 +149,8 @@ en:
             oauth_parameter_cannot_be_secure: oauth parameter cannot be set to be
               secure.
             invalid_url: '%{field} must be a valid URL, got "%{value}".'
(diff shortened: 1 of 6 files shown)


mmassaki (approved):
Thank you for addressing my comments.

JeanMarcGoepfert (approved):

PR #396, opened 2025-09-24
[CRYSTAL-514] Add a flag for the gating validation of cov2 requirements

zendesk-punavwalke:
💐

/cc @zendesk/wattle

### Description

The PR includes changes related to addition of flag to prevent validation of cov2 requirements when feature flag is disabled in ZAM.
- Extend Package#validate/validate! with custom_objects_v2_requirements parameter
- Update Requirements.call to conditionally validate COV2 based on opts flag
- Add thorough test coverage for both enabled/disabled validation scenarios

### References
[Link to a JIRA or GitHub issue here if relevant](https://zendesk.atlassian.net/browse/CRYSTAL-514)

#### Before merging this PR

### Risks
* [RUNTIME] Can this change affect apps rendering for a user?
Yes, it is called during app zip validation
* [HIGH | medium | low] What features does this touch?
Low: This change is called during zip validation

### DIFF
--- a/lib/zendesk_apps_support/package.rb
+++ b/lib/zendesk_apps_support/package.rb
@@ -31,7 +31,8 @@ module ZendeskAppsSupport
       @warnings = []
     end
 
-    def validate(marketplace: true, skip_marketplace_translations: false, error_on_password_parameter: false)
+    def validate(marketplace: true, skip_marketplace_translations: false, error_on_password_parameter: false,
+                 validate_custom_objects_v2: false)
       errors = []
       errors << Validations::Manifest.call(self, error_on_password_parameter: error_on_password_parameter)
 
@@ -39,7 +40,7 @@ module ZendeskAppsSupport
         errors << Validations::Marketplace.call(self) if marketplace
         errors << Validations::Source.call(self)
         errors << Validations::Translations.call(self, skip_marketplace_translations: skip_marketplace_translations)
-        errors << Validations::Requirements.call(self)
+        errors << Validations::Requirements.call(self, validate_custom_objects_v2:)
 
         # only adds warnings
         Validations::SecureSettings.call(self)
@@ -61,9 +62,14 @@ module ZendeskAppsSupport
       errors.flatten.compact
     end
 
-    def validate!(marketplace: true, skip_marketplace_translations: false, error_on_password_parameter: false)
-      errors = validate(marketplace: marketplace, skip_marketplace_translations: skip_marketplace_translations, error_on_password_parameter: error_on_password_parameter)
+    def validate!(marketplace: true, skip_marketplace_translations: false, error_on_password_parameter: false,
+                  validate_custom_objects_v2: false)
+      errors = validate(marketplace: marketplace,
+                        skip_marketplace_translations: skip_marketplace_translations,
+                        error_on_password_parameter: error_on_password_parameter,
+                        validate_custom_objects_v2:)
       raise errors.first if errors.any?
+
       true
     end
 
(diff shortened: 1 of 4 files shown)


lib/zendesk_apps_support/validations/requirements.rb
```diff
@@ -31,7 +31,7 @@ def call(package)
             errors << invalid_custom_objects(requirements)
             errors << invalid_webhooks(requirements)
             errors << invalid_target_types(requirements)
-            errors << validate_custom_objects_v2_requirements(requirements)
+            errors << validate_custom_objects_v2_requirements(requirements, opts)
             errors << missing_required_fields(requirements)
```
zendesk-punavwalke, line 35: Avoided to prevent call to `validate_custom_objects_v2_requirements` because of `rubocop Cyclomatic complexity`. Instead I have added a early return from `validate_custom_objects_v2_requirements` when cov2_validation is not allowed.

vaibhavmdhoke (approved):
Added comments for minor changes, LGTM!

spec/validations/requirements_spec.rb
```diff
@@ -362,63 +362,94 @@
...
+          'object_fields' => [{ 
+            'key' => 'field_1', 'type' => 'text', 'title' => 'Field 1', 'object_key' => 'object_1' 
+          }],
+          'object_triggers' => []
+        }
+      )
+    end
+
+    context 'when enable_custom_objects_v2_validation is false (default)' do
+      context 'with no custom_objects_v2 requirements' do
```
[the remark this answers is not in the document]
vaibhavmdhoke: ok, if it's covered not needed

lib/zendesk_apps_support/validations/requirements.rb
```diff
@@ -7,7 +7,7 @@ module Requirements
       MAX_CUSTOM_OBJECTS_REQUIREMENTS = 50
 
       class << self
-        def call(package)
+        def call(package, opts = {})
```
mmassaki, line 10: Could we make the new arg explicit instead of a generic option (e.g. `validate_custom_objects_v2: false`)? That would make it easier to remove later since this should be a temporary change.


```suggestion
        def call(package, validate_custom_objects_v2: false)
```

lib/zendesk_apps_support/package.rb
```diff
@@ -31,15 +31,19 @@ def initialize(dir, is_cached = true)
       @warnings = []
     end
 
-    def validate(marketplace: true, skip_marketplace_translations: false, error_on_password_parameter: false)
+    def validate(marketplace: true, skip_marketplace_translations: false, error_on_password_parameter: false,
+                 enable_custom_objects_v2_validation: false)
```
mmassaki, line 35: nit: how about renaming the arg `validate_custom_objects_v2_requirements` or `validate_custom_objects_v2` to be shorter?

lib/zendesk_apps_support/validations/requirements.rb
zendesk-punavwalke: Yes,
`errors << validate_custom_objects_v2_requirements(requirements) if validate_custom_objects_v2`
is a rubocop violation because of conditions at top. 

Extracting above conditions to helper method would get rid of this warning.However, I intentionally avoided making that change to minimize modifications to existing code.  
@mmassaki - I'm open to making this refactoring change if you think it's appropriate.

lib/zendesk_apps_support/package.rb
mmassaki: I'd recommend avoiding acronyms when possible. `validate_custom_objects_v2` seems a better option IMHO.

mmassaki (approved):

PR #422, opened 2026-04-20
[CRYSTAL-923] Remove conditional cov2 requirements validation

zendesk-punavwalke:
/cc @zendesk/wattle

### Description

The `validate_custom_objects_v2` parameter was used to control whether COV2 requirements validation runs. This was gated by the `apps_cov2_processing` Arturo in `zendesk_app_market`, which is now fully rolled out and stable.

This PR removes the conditional flag so that COV2 validation always runs whenever the `custom_objects_v2` key exists in `requirements.json`.

**Changes:**
- `Package#validate` no longer reads or passes `validate_custom_objects_v2`
- `Validations::Requirements.call` no longer accepts the keyword argument
- `validate_custom_objects_v2_requirements` now only checks if the key is present (no flag check)
- Tests updated to remove flag-based branching (enabled/disabled contexts removed)

Callers that still pass `validate_custom_objects_v2` in their options hash will not break — `Package#validate` uses `options = {}` so unknown keys are silently ignored.

### References

JIRA: https://zendesk.atlassian.net/browse/CRYSTAL-762

#### Before merging this PR

### Risks

- [low] Apps that have `custom_objects_v2` in their `requirements.json` will now always be validated, regardless of the caller passing the flag. This is the intended behavior since the feature is stable and fully rolled out.

### DIFF
--- a/lib/zendesk_apps_support/validations/requirements.rb
+++ b/lib/zendesk_apps_support/validations/requirements.rb
@@ -7,7 +7,7 @@ module ZendeskAppsSupport
       MAX_CUSTOM_OBJECTS_REQUIREMENTS = 50
 
       class << self
-        def call(package, validate_custom_objects_v2: false)
+        def call(package)
           if package.manifest.requirements_only? && !package.has_requirements?
             return [ValidationError.new(:missing_requirements)]
           elsif !supports_requirements(package) && package.has_requirements?
@@ -31,7 +31,7 @@ module ZendeskAppsSupport
             errors << invalid_custom_objects(requirements)
             errors << invalid_webhooks(requirements)
             errors << invalid_target_types(requirements)
-            errors << validate_custom_objects_v2_requirements(requirements, validate_custom_objects_v2:)
+            errors << validate_custom_objects_v2_requirements(requirements)
             errors << missing_required_fields(requirements)
             errors.flatten!
             errors.compact!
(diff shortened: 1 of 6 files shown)


lib/zendesk_apps_support/validations/requirements.rb
```diff
@@ -7,7 +7,7 @@ module Requirements
       MAX_CUSTOM_OBJECTS_REQUIREMENTS = 50
 
       class << self
-        def call(package, validate_custom_objects_v2: false)
+        def call(package)
```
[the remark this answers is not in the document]
zendesk-punavwalke: Requirements.call is an internal method — only called from `Package#validate` within the same gem. External callers use `Package#validate(options)` which accepts a hash, so passing extra keys is not breaking change.

vaibhavmdhoke (dismissed):

spec/validations/requirements_spec.rb
```diff
@@ -390,66 +390,30 @@
...
+      let(:requirements_string) { valid_objects_requirements }
 
-        it 'calls validate_custom_objects_v2_requirements and validates successfully' do
-          expect(ZendeskAppsSupport::Validations::Requirements)
-            .to receive(:validate_custom_objects_v2_requirements).and_call_original
-          expect(ZendeskAppsSupport::Validations::CustomObjectsV2).to receive(:call).and_return([])
-          expect(errors).to be_empty
-        end
+      it 'validates successfully' do
+        expect(ZendeskAppsSupport::Validations::CustomObjectsV2).to receive(:call).and_return([])
```
mmassaki: Calling `#and_return([])` here mocks the method to return an empty response, so it's not actually validating the requirements. Consider changing it to `#and_call_original`, otherwise the test will never fail.


```suggestion
        expect(ZendeskAppsSupport::Validations::CustomObjectsV2).to receive(:call).and_call_original
```

zendesk-punavwalke: Thanks for pointing this out! Updated to `.and_call_original`

mmassaki (approved):

PR #428, opened 2026-06-19
feat: Add object_types support for cov2_records_sidebar location

vijaytembugade:
💐

/cc @zendesk/wattle

### Description

- Add new `custom_object_record_sidebar` location (id: 27) for the Support product, enabling apps to render in a custom objects v2 records sidebar. The location is orderable, collapsible, and visible.
- Introduce `objectTypes` location option scoped exclusively to `custom_object_record_sidebar`, allowing apps to declare which custom object types the sidebar should render for. `objectTypes` is **required** for `custom_object_record_sidebar` — omitting it raises `:object_types_required`.
- Add manifest validations for `objectTypes`: must be a non-empty array of non-empty strings; rejected on any location other than `custom_object_record_sidebar` when the flag is enabled. Edge cases covered include whitespace-only strings, nil values, and numeric entries.
- Gate the `custom_object_record_sidebar` location behind a `validate_custom_object_record_sidebar_location` option passed through `Package#validate` → `Validations::Manifest.call`. When the flag is off (default), the location is reported as invalid. **Known gap:** when the flag is off, `objectTypes` on supported locations (e.g., `ticket_sidebar`) is silently ignored rather than rejected — a follow-up is needed to raise `:object_types_not_supported` regardless of the flag state.
- Add locale strings and translator-friendly translations for **five** new `objectTypes` validation errors:
  1. `object_types_must_be_array` — value is not an array
  2. `object_types_empty` — array is present but empty
  3. `object_types_invalid_entries` — array contains non-string, empty, or whitespace-only entries
  4. `object_types_not_supported` — `objectTypes` specified on a location that does not support it
  5. `object_types_required` — `objectTypes` omitted on `custom_object_record_sidebar`
- Add comprehensive specs covering location options parsing, manifest parsing, and all validation rules including edge cases, multi-location manifests, and default flag behaviour.

### References

- [CRYSTAL-1381](https://zendesk.atlassian.net/browse/CRYSTAL-1381)

### Screenshots & GIFs

- N/A

#### Before merging this PR

### Risks

- Medium: `objectTypes` is **required** for `custom_object_record_sidebar`, making this a stricter contract than "validated only when present." There are five validation error paths (not four as originally noted — includes `object_types_required`). When the flag is disabled (default), `objectTypes` on supported locations (e.g., `ticket_sidebar`) is silently ignored rather than rejected — apps could ship manifests containing `objectTypes` on non-target locations without error until the flag is turned on, at which point those manifests would start failing validation. Existing locations and manifest parsing are unaffected. When the flag is off, `custom_object_record_sidebar` is treated as an invalid location, so no apps can use it until explicitly enabled.

[CRYSTAL-1381]: https://zendesk.atlassian.net/browse/CRYSTAL-1381

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -305,13 +305,13 @@ en:
             field_cannot_be_empty: "%{field} cannot be empty."
             field_contains_invalid_values: "%{field} contains invalid values: %{values}."
             field_contains_invalid_keys: "%{field} contains invalid keys: %{keys}."
-            object_types_must_be_array: object_types must be an array of strings for
+            object_types_must_be_array: objectTypes must be an array of strings for
               the %{location} location, but it was %{value}.
-            object_types_empty: object_types cannot be empty for the %{location} location.
+            object_types_empty: objectTypes cannot be empty for the %{location} location.
               Specify at least one object type.
-            object_types_invalid_entries: object_types must contain only non-empty
+            object_types_invalid_entries: objectTypes must contain only non-empty
               strings for the %{location} location.
-            object_types_not_supported: object_types is not supported for the %{location}
+            object_types_not_supported: objectTypes is not supported for the %{location}
               location.
         warning:
           app_build:
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -268,6 +268,9 @@ en:
                 do not match products in translations (%{translation_products})
               insecure_token_parameter_in_manifest: 'Make sure to set secure to true
                 when using keys in Settings. Learn more: %{link}'
+              password_parameter_deprecated: 'Password parameter type is deprecated
+                and will not be accepted in the future. Use Basic Access Authentication
+                instead. Learn more: %{link}.'
               default_secure_or_hidden_parameter_in_manifest: Default values for secure
                 or hidden parameters are not stored securely. Be sure to review them
                 and confirm they do not contain sensitive data
@@ -295,11 +298,21 @@ en:
             oauth_parameter_cannot_be_secure: oauth parameter cannot be set to be
               secure.
             invalid_url: '%{field} must be a valid URL, got "%{value}".'
+            password_parameter_type_deprecated: 'Password parameter type can no longer
+              be used. Use Secure settings instead. Learn more: %{link}.'
             field_requires_secure_parameter: "%{field} cannot be defined in a non-secure
               parameter."
             field_cannot_be_empty: "%{field} cannot be empty."
             field_contains_invalid_values: "%{field} contains invalid values: %{values}."
             field_contains_invalid_keys: "%{field} contains invalid keys: %{keys}."
+            object_types_must_be_array: object_types must be an array of strings for
+              the %{location} location, but it was %{value}.
+            object_types_empty: object_types cannot be empty for the %{location} location.
+              Specify at least one object type.
+            object_types_invalid_entries: object_types must contain only non-empty
+              strings for the %{location} location.
+            object_types_not_supported: object_types is not supported for the %{location}
+              location.
         warning:
           app_build:
             deprecated_version: You are targeting a deprecated version of the framework.
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -300,6 +300,14 @@ en:
             field_cannot_be_empty: "%{field} cannot be empty."
             field_contains_invalid_values: "%{field} contains invalid values: %{values}."
             field_contains_invalid_keys: "%{field} contains invalid keys: %{keys}."
+            object_types_not_supported: object_types is not supported for the %{location}
+              location. It can only be used with cov2_records_sidebar.
+            object_types_must_be_array: object_types must be an array of strings for
+              the %{location} location, but it was %{value}.
+            object_types_empty: object_types cannot be empty for the %{location} location.
+              Specify at least one object type.
+            object_types_invalid_entries: object_types must contain only non-empty strings
+              for the %{location} location.
         warning:
           app_build:
             deprecated_version: You are targeting a deprecated version of the framework.
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -313,6 +313,7 @@ en:
               strings for the %{location} location.
             object_types_not_supported: objectTypes is not supported for the %{location}
               location.
+            object_types_required: objectTypes is required for the %{location} location.
         warning:
           app_build:
             deprecated_version: You are targeting a deprecated version of the framework.
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -300,14 +300,6 @@ en:
             field_cannot_be_empty: "%{field} cannot be empty."
             field_contains_invalid_values: "%{field} contains invalid values: %{values}."
             field_contains_invalid_keys: "%{field} contains invalid keys: %{keys}."
-            object_types_not_supported: object_types is not supported for the %{location}
-              location. It can only be used with cov2_records_sidebar.
-            object_types_must_be_array: object_types must be an array of strings for
-              the %{location} location, but it was %{value}.
-            object_types_empty: object_types cannot be empty for the %{location} location.
-              Specify at least one object type.
-            object_types_invalid_entries: object_types must contain only non-empty strings
-              for the %{location} location.
         warning:
           app_build:
             deprecated_version: You are targeting a deprecated version of the framework.
--- a/lib/zendesk_apps_support/location.rb
+++ b/lib/zendesk_apps_support/location.rb
@@ -65,7 +65,8 @@ module ZendeskAppsSupport
       Location.new(id: 23, name: 'call_log_editor', product_code: Product::SELL.code, visible: true),
       Location.new(id: 24, name: 'email_editor', product_code: Product::SELL.code, visible: true),
       Location.new(id: 25, name: 'top_bar', product_code: Product::SELL.code, visible: true),
-      Location.new(id: 26, name: 'visit_editor', product_code: Product::SELL.code, visible: true)
+      Location.new(id: 26, name: 'visit_editor', product_code: Product::SELL.code, visible: true),
+      Location.new(id: 27, name: 'cov2_records_sidebar', product_code: Product::SUPPORT.code, visible: true)
     ].freeze
   end
 end
--- a/lib/zendesk_apps_support/location.rb
+++ b/lib/zendesk_apps_support/location.rb
@@ -68,7 +68,7 @@ module ZendeskAppsSupport
       Location.new(id: 24, name: 'email_editor', product_code: Product::SELL.code, visible: true),
       Location.new(id: 25, name: 'top_bar', product_code: Product::SELL.code, visible: true),
       Location.new(id: 26, name: 'visit_editor', product_code: Product::SELL.code, visible: true),
-      Location.new(id: 27, orderable: true, collapsible: true, name: 'custom_object_record_sidebar',
+      Location.new(id: 27, orderable: true, collapsible: true, name: CUSTOM_OBJECT_RECORD_SIDEBAR_LOCATION,
                    product_code: Product::SUPPORT.code, visible: true)
     ].freeze
   end
--- a/lib/zendesk_apps_support/location.rb
+++ b/lib/zendesk_apps_support/location.rb
@@ -5,6 +5,8 @@ module ZendeskAppsSupport
     extend ZendeskAppsSupport::Finders
     attr_reader :id, :name, :orderable, :collapsible, :visible, :product_code, :v2_only
 
+    CUSTOM_OBJECT_RECORD_SIDEBAR_LOCATION = 'custom_object_record_sidebar'
+
     def self.unique_ids
       @ids ||= Set.new
     end
@@ -24,8 +26,6 @@ module ZendeskAppsSupport
     def product
       Product.find_by(code: product_code)
     end
-
-    OBJECT_TYPES_LOCATION = 'cov2_records_sidebar'
 
     def self.all
       LOCATIONS_AVAILABLE
@@ -68,7 +68,7 @@ module ZendeskAppsSupport
       Location.new(id: 24, name: 'email_editor', product_code: Product::SELL.code, visible: true),
       Location.new(id: 25, name: 'top_bar', product_code: Product::SELL.code, visible: true),
       Location.new(id: 26, name: 'visit_editor', product_code: Product::SELL.code, visible: true),
-      Location.new(id: 27, orderable: true, collapsible: true, name: 'cov2_records_sidebar',
+      Location.new(id: 27, orderable: true, collapsible: true, name: 'custom_object_record_sidebar',
                    product_code: Product::SUPPORT.code, visible: true)
     ].freeze
   end
--- a/lib/zendesk_apps_support/location.rb
+++ b/lib/zendesk_apps_support/location.rb
@@ -25,6 +25,8 @@ module ZendeskAppsSupport
       Product.find_by(code: product_code)
     end
 
+    OBJECT_TYPES_LOCATION = 'cov2_records_sidebar'
+
     def self.all
       LOCATIONS_AVAILABLE
     end
@@ -66,7 +68,8 @@ module ZendeskAppsSupport
       Location.new(id: 24, name: 'email_editor', product_code: Product::SELL.code, visible: true),
       Location.new(id: 25, name: 'top_bar', product_code: Product::SELL.code, visible: true),
       Location.new(id: 26, name: 'visit_editor', product_code: Product::SELL.code, visible: true),
-      Location.new(id: 27, name: 'cov2_records_sidebar', product_code: Product::SUPPORT.code, visible: true)
+      Location.new(id: 27, orderable: true, collapsible: true, name: 'cov2_records_sidebar',
+                   product_code: Product::SUPPORT.code, visible: true)
     ].freeze
   end
 end
--- a/lib/zendesk_apps_support/manifest.rb
+++ b/lib/zendesk_apps_support/manifest.rb
@@ -98,7 +98,7 @@ module ZendeskAppsSupport
     end
 
     def locations_for_product(host)
-      locations.key?(host) ? locations[host].keys : []
+      locations.key?(host) ? locations[host].keys.map(&:to_s) : []
     end
 
     def unknown_hosts
--- a/lib/zendesk_apps_support/manifest.rb
+++ b/lib/zendesk_apps_support/manifest.rb
@@ -97,10 +97,6 @@ module ZendeskAppsSupport
       end
     end
 
-    def locations_for_product(host)
-      locations.key?(host) ? locations[host].keys.map(&:to_s) : []
-    end
-
     def unknown_hosts
       @unknown_hosts ||=
         @used_hosts - Product::PRODUCTS_AVAILABLE.flat_map { |p| [p.name, p.legacy_name] }
--- a/lib/zendesk_apps_support/manifest.rb
+++ b/lib/zendesk_apps_support/manifest.rb
@@ -97,6 +97,10 @@ module ZendeskAppsSupport
       end
     end
 
+    def locations_for_product(host)
+      locations.key?(host) ? locations[host].keys : []
+    end
+
     def unknown_hosts
       @unknown_hosts ||=
         @used_hosts - Product::PRODUCTS_AVAILABLE.flat_map { |p| [p.name, p.legacy_name] }
--- a/lib/zendesk_apps_support/manifest/location_options.rb
+++ b/lib/zendesk_apps_support/manifest/location_options.rb
@@ -11,7 +11,7 @@ module ZendeskAppsSupport
         signed: 'signed',
         url: 'url',
         size: 'size',
-        object_types: 'object_types'
+        object_types: 'objectTypes'
       }.freeze
 
       OBJECT_TYPES_LOCATION = 'cov2_records_sidebar'
--- a/lib/zendesk_apps_support/manifest/location_options.rb
+++ b/lib/zendesk_apps_support/manifest/location_options.rb
@@ -10,8 +10,11 @@ module ZendeskAppsSupport
         flexible: 'flexible',
         signed: 'signed',
         url: 'url',
-        size: 'size'
+        size: 'size',
+        object_types: 'object_types'
       }.freeze
+
+      OBJECT_TYPES_LOCATION = 'cov2_records_sidebar'
 
       attr_reader :location
       attr_reader(*RUBY_TO_JSON.keys)
@@ -31,6 +34,7 @@ module ZendeskAppsSupport
         @auto_load = options.fetch('autoLoad', true)
         @auto_hide ||= false
         @signed ||= false
+        @object_types = nil unless @location&.name == OBJECT_TYPES_LOCATION
       end
     end
   end
--- a/lib/zendesk_apps_support/manifest/location_options.rb
+++ b/lib/zendesk_apps_support/manifest/location_options.rb
@@ -14,8 +14,6 @@ module ZendeskAppsSupport
         object_types: 'objectTypes'
       }.freeze
 
-      OBJECT_TYPES_LOCATION = 'cov2_records_sidebar'
-
       attr_reader :location
       attr_reader(*RUBY_TO_JSON.keys)
 
@@ -34,7 +32,6 @@ module ZendeskAppsSupport
         @auto_load = options.fetch('autoLoad', true)
         @auto_hide ||= false
         @signed ||= false
-        @object_types = nil unless @location&.name == OBJECT_TYPES_LOCATION
       end
     end
   end
--- a/lib/zendesk_apps_support/package.rb
+++ b/lib/zendesk_apps_support/package.rb
@@ -12,9 +12,9 @@ module ZendeskAppsSupport
     MANIFEST_FILENAME = 'manifest.json'
     REQUIREMENTS_FILENAME = 'requirements.json'
 
-    DEFAULT_LAYOUT = Erubis::Eruby.new(File.read(File.expand_path('assets/default_template.html.erb', __dir__)))
-    DEFAULT_SCSS   = File.read(File.expand_path('assets/default_styles.scss', __dir__))
-    SRC_TEMPLATE   = Erubis::Eruby.new(File.read(File.expand_path('assets/src.js.erb', __dir__)))
+    DEFAULT_LAYOUT = Erubis::Eruby.new(File.read(File.expand_path('../assets/default_template.html.erb', __FILE__)))
+    DEFAULT_SCSS   = File.read(File.expand_path('../assets/default_styles.scss', __FILE__))
+    SRC_TEMPLATE   = Erubis::Eruby.new(File.read(File.expand_path('../assets/src.js.erb', __FILE__)))
 
     LOCATIONS_WITH_ICONS_PER_PRODUCT = {
       Product::SUPPORT => %w[top_bar nav_bar system_top_bar ticket_editor].freeze,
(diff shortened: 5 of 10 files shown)


lib/zendesk_apps_support/manifest/location_options.rb
```diff
@@ -10,9 +10,12 @@ class LocationOptions
         flexible: 'flexible',
         signed: 'signed',
         url: 'url',
-        size: 'size'
+        size: 'size',
+        object_types: 'object_types'
```
zendesk-punavwalke: value should be in camel case `objectTypes` and not `object_types`([ref)](https://github.com/zendesk/zendesk_apps_support/blob/322f41d3eab28109df20630b1113977b8c195273/lib/zendesk_apps_support/manifest/location_options.rb#L8)

config/locales/en.yml
```diff
@@ -300,6 +300,14 @@ en:
...
             field_contains_invalid_values: "%{field} contains invalid values: %{values}."
             field_contains_invalid_keys: "%{field} contains invalid keys: %{keys}."
+            object_types_not_supported: object_types is not supported for the %{location}
+              location. It can only be used with cov2_records_sidebar.
+            object_types_must_be_array: object_types must be an array of strings for
+              the %{location} location, but it was %{value}.
+            object_types_empty: object_types cannot be empty for the %{location} location.
+              Specify at least one object type.
+            object_types_invalid_entries: object_types must contain only non-empty strings
+              for the %{location} location.
```
zendesk-punavwalke: create a separate PR for just translations

This is an auto generated file, do not directly modify it for adding the translations

This is [where](https://github.com/zendesk/zendesk_apps_support/blob/main/config/locales/translations/zendesk_apps_support.yml) you have add new translations

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -60,7 +60,8 @@ def validate_location(package)
             missing_location_error(package),
             invalid_location_error(package),
             invalid_v1_location(package),
-            location_framework_mismatch(manifest)
+            location_framework_mismatch(manifest),
+            validate_object_types(manifest)
           ]
```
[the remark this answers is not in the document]
vijaytembugade: Updated : 
validate_object_types does not return an Array now. It returns either a single ValidationError (via early return) or nil (the implicit return value of each when no guard triggers). This is consistent with other validators like location_framework_mismatch, invalid_v1_location, and ban_no_template.

lib/zendesk_apps_support/manifest/location_options.rb
```diff
@@ -31,6 +34,7 @@ def initialize(location, options)
         @auto_load = options.fetch('autoLoad', true)
         @auto_hide ||= false
         @signed ||= false
+        @object_types = nil unless @location&.name == OBJECT_TYPES_LOCATION
```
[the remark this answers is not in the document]
vijaytembugade: This is intentional — we chose option (2): silently ignore object_types for unsupported locations.
Updated validate_object_types function too.

lib/zendesk_apps_support/manifest/location_options.rb
vijaytembugade: correct. Updated this.

config/locales/en.yml
vijaytembugade: I have updated the respective translation in proper files. 
If I created new PR I might need to create multiple PRs for version bump. So, keeping it in same.

zendesk-punavwalke (changes requested):
Please gate this change behind an Arturo flag so `cov2_records_sidebar` and `objectTypes` validation are active only when the flag is enabled.

config/locales/en.yml
zendesk-punavwalke: Also, please run` bundle exec rake i18n:standardize` after updating the [source translation file](https://github.com/zendesk/zendesk_apps_support/blob/main/config/locales/translations/zendesk_apps_support.yml).
You can see this noted [here](https://github.com/zendesk/zendesk_apps_support/blob/ccc9a88e0fb0dfc1bad84936a83081d8bbf4ffad/config/locales/en.yml).
This is needed because `en.yml `is generated output used at runtime.

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -379,6 +389,42 @@ def invalid_location_flexible_error(location_options)
...
+            end
+
+            non_strings = object_types.reject { |t| t.is_a?(String) && !t.strip.empty? }
+            if non_strings.any?
+              return ValidationError.new(:object_types_invalid_entries,
+                                         location: location_name)
+            end
+          end
+          nil
+        end
```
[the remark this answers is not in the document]
vijaytembugade: https://github.com/zendesk/zendesk_apps_support/pull/428#discussion_r3441396017

I was doing the same, But copilot itself suggested me to not return explicit array.

config/locales/en.yml
```diff
@@ -268,6 +268,9 @@ en:
                 do not match products in translations (%{translation_products})
               insecure_token_parameter_in_manifest: 'Make sure to set secure to true
                 when using keys in Settings. Learn more: %{link}'
+              password_parameter_deprecated: 'Password parameter type is deprecated
+                and will not be accepted in the future. Use Basic Access Authentication
+                instead. Learn more: %{link}.'
```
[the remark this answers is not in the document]
vijaytembugade: This is autogenerated.

config/locales/en.yml
```diff
@@ -295,11 +298,21 @@ en:
             oauth_parameter_cannot_be_secure: oauth parameter cannot be set to be
               secure.
             invalid_url: '%{field} must be a valid URL, got "%{value}".'
+            password_parameter_type_deprecated: 'Password parameter type can no longer
+              be used. Use Secure settings instead. Learn more: %{link}.'
```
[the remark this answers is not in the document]
vijaytembugade: this is autogenerated.

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -14,15 +14,15 @@ module Manifest
       SECURE_PARAM_SCOPES = %w[header body url jwt_secret_key jwt_claim basic_auth_username basic_auth_password].freeze
 
       class << self
-        def call(package)
+        def call(package, opts = {})
```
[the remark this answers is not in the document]
vijaytembugade: Consistency with the rest of the codebase -- Package#validate already uses the same options = {} / opts.fetch(:key, default) pattern (see package.rb), and the opts hash is threaded through the entire call -> collate_manifest_errors -> validate_location -> invalid_location_error / validate_object_types chain.

Switching only this file to keyword args would create an inconsistent interface at the boundary where Package#validate passes options in.

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -329,6 +333,12 @@ def invalid_location_error(package)
 
           Product::PRODUCTS_AVAILABLE.each do |product|
             invalid_locations = package.manifest.unknown_locations(product.name)
+            unless is_cov2_sidebar_app_enabled
+              requested_locations = package.manifest.locations_for_product(product.name)
+              if requested_locations.include?(cov2_location)
+                invalid_locations << cov2_location
+              end
```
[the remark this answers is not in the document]
vijaytembugade: https://github.com/zendesk/zendesk_apps_support/pull/428/commits/4f64407956d18a1d536680fb68f7eda396aae7b0

handled in this commit

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -379,6 +389,47 @@ def invalid_location_flexible_error(location_options)
...
+
+            if location_name != cov2_location
+              return ValidationError.new(:object_types_not_supported,
+                                         location: location_name)
+            end
+
+            if location_name == cov2_location && !is_cov2_sidebar_app_enabled
+              return ValidationError.new(:object_types_not_supported,
+                                         location: location_name)
+            end
```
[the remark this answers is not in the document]
vijaytembugade: description is updated.

config/locales/en.yml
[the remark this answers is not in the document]
vijaytembugade: This is autogenerated.

config/locales/en.yml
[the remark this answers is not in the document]
vijaytembugade: this is also autogenerated.

lib/zendesk_apps_support/manifest.rb
```diff
@@ -97,6 +97,10 @@ def unknown_locations(host)
       end
     end
 
+    def locations_for_product(host)
+      locations.key?(host) ? locations[host].keys.map(&:to_s) : []
+    end
```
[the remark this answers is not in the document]
vijaytembugade: This is safe to dismiss. The locations accessor is always initialized to a Hash by set_locations_and_hosts (even the nil/else branch returns { 'support' => {} }), so it can never be nil. The keys are always strings (sourced from the JSON manifest), and host is always passed as product.name which is also a string — so there's no symbol/string mismatch. This is the same pattern used by unknown_locations directly above, which has been stable in this codebase for a long time.

lib/zendesk_apps_support/manifest/location_options.rb
```diff
@@ -10,9 +10,12 @@ class LocationOptions
         flexible: 'flexible',
         signed: 'signed',
         url: 'url',
-        size: 'size'
+        size: 'size',
+        object_types: 'objectTypes'
       }.freeze
 
+      OBJECT_TYPES_LOCATION = 'cov2_records_sidebar'
```
zendesk-punavwalke: Can we move `OBJECT_TYPES_LOCATION` from this file to `location.rb` (or shared location constants) to keep separation of concerns clean? Its value is a location type/identifier, not a location-options parsing concern.

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -329,6 +333,12 @@ def invalid_location_error(package)
...
           Product::PRODUCTS_AVAILABLE.each do |product|
             invalid_locations = package.manifest.unknown_locations(product.name)
+            unless is_cov2_sidebar_app_enabled
+              requested_locations = package.manifest.locations_for_product(product.name)
+              if requested_locations.include?(cov2_location)
+                invalid_locations << cov2_location
+              end
+            end
             next if invalid_locations.empty?
             errors << ValidationError.new(:invalid_location,
```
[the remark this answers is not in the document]
vijaytembugade: This is intentional. When the feature flag is off, both errors are valid and provide distinct information to the app developer:

:invalid_location tells them cov2_records_sidebar isn't a recognized location
:object_types_not_supported tells them the objectTypes option isn't available

lib/zendesk_apps_support/location.rb
```diff
@@ -65,7 +65,8 @@ def self.all
       Location.new(id: 23, name: 'call_log_editor', product_code: Product::SELL.code, visible: true),
       Location.new(id: 24, name: 'email_editor', product_code: Product::SELL.code, visible: true),
       Location.new(id: 25, name: 'top_bar', product_code: Product::SELL.code, visible: true),
-      Location.new(id: 26, name: 'visit_editor', product_code: Product::SELL.code, visible: true)
+      Location.new(id: 26, name: 'visit_editor', product_code: Product::SELL.code, visible: true),
+      Location.new(id: 27, name: 'cov2_records_sidebar', product_code: Product::SUPPORT.code, visible: true)
```
zendesk-punavwalke: For `cov2_records_sidebar,` is omission of `orderable: true and collapsible: true` intentional?

lib/zendesk_apps_support/package.rb
```diff
@@ -34,9 +34,10 @@ def initialize(dir, is_cached = true)
     def validate(options = {})
       marketplace = options.fetch(:marketplace, true)
       skip_marketplace_translations = options.fetch(:skip_marketplace_translations, false)
+      is_cov2_sidebar_app_enabled = options.fetch(:is_cov2_sidebar_app_enabled, false)
```
zendesk-punavwalke: Rename `is_cov2_sidebar_app_enabled` to `validate_cov2_records_sidebar_location `(or something more meaningful)? Because this flag is checking whether validation for cov2_records_sidebar/objectTypes runs, not whether a sidebar/app is enabled.

lib/zendesk_apps_support/manifest.rb
```diff
@@ -97,6 +97,10 @@ def unknown_locations(host)
       end
     end
 
+    def locations_for_product(host)
```
zendesk-punavwalke: Gate `cov2_records_sidebar` at the location source (location.rb) when the flag is off, instead of keeping it globally available and marking it invalid later in manifest validation.
This avoids workaround logic here.

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -329,6 +333,12 @@ def invalid_location_error(package)
 
           Product::PRODUCTS_AVAILABLE.each do |product|
             invalid_locations = package.manifest.unknown_locations(product.name)
+            unless is_cov2_sidebar_app_enabled
```
zendesk-punavwalke: If `cov2_records_sidebar` is gated/filtered at the location source as mentioned [here](https://github.com/zendesk/zendesk_apps_support/pull/428/changes#r3466059395), this extra invalidation logic should not be needed.
Existing unknown_locations -> invalid_location handling will cover the flag-off path.

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -315,6 +316,7 @@ def missing_location_error(package)
 
         def invalid_location_error(package)
           errors = []
+
```
zendesk-punavwalke, line 323: nit: remove extra space

lib/zendesk_apps_support/location.rb
```diff
@@ -25,8 +29,12 @@ def product
       Product.find_by(code: product_code)
     end
 
+    OBJECT_TYPES_LOCATION = 'cov2_records_sidebar'
```
zendesk-punavwalke: Rename this constant to `COV2_RECORDS_SIDEBAR_LOCATION`
It stores a location identifier, not an `objectTypes` concept.

lib/zendesk_apps_support/location.rb
```diff
@@ -5,6 +5,10 @@ class Location
     extend ZendeskAppsSupport::Finders
     attr_reader :id, :name, :orderable, :collapsible, :visible, :product_code, :v2_only
 
+    class << self
```
zendesk-punavwalke: Remove this class getter/setter; it introduces global mutable state.
The Arturo flag value should be passed through method options from [here](https://github.com/zendesk/zendesk_apps_support/blob/39aa6fc77d3a14f398267bd9a3e0df707c9a61a8/lib/zendesk_apps_support/package.rb#L41)

lib/zendesk_apps_support/package.rb
```diff
@@ -34,6 +34,8 @@ def initialize(dir, is_cached = true)
     def validate(options = {})
       marketplace = options.fetch(:marketplace, true)
       skip_marketplace_translations = options.fetch(:skip_marketplace_translations, false)
+      validate_cov2_records_sidebar_location = options.fetch(:validate_cov2_records_sidebar_location, false)
+      Location.validate_cov2_records_sidebar_location = validate_cov2_records_sidebar_location
```
zendesk-punavwalke: This creates global mutable state. Keep the flag in method options and pass it through the validation methods per call.


```suggestion
      errors << Validations::Manifest.call(
    self,
    validate_cov2_records_sidebar_location: validate_cov2_records_sidebar_location
  )
```

lib/zendesk_apps_support/location.rb
vijaytembugade: @zendesk-punavwalke 
https://github.com/zendesk/zendesk_apps_support/pull/428#discussion_r3466171256, in this comment you mentioned it to moved here

zendesk-punavwalke:
@vijaytembugade Suggestion: Revert the class-level flag state approach in location.rb (`attr_accessor + assignment from package.rb`).
Pass the Arturo-derived flag from `package.rb `to manifest validation via method options, and apply flag-based filtering there to mark `cov2_records_sidebar` as invalid when Arturo is off.

lib/zendesk_apps_support/location.rb
```diff
@@ -65,7 +73,9 @@ def self.all
       Location.new(id: 23, name: 'call_log_editor', product_code: Product::SELL.code, visible: true),
       Location.new(id: 24, name: 'email_editor', product_code: Product::SELL.code, visible: true),
       Location.new(id: 25, name: 'top_bar', product_code: Product::SELL.code, visible: true),
-      Location.new(id: 26, name: 'visit_editor', product_code: Product::SELL.code, visible: true)
+      Location.new(id: 26, name: 'visit_editor', product_code: Product::SELL.code, visible: true),
+      Location.new(id: 27, orderable: true, collapsible: true, name: 'cov2_records_sidebar',
```
zendesk-punavwalke: name this location as `cov2_record_sidebar` instead of `cov2_records_sidebar` just to be consistent with other locations

zendesk-punavwalke: Updating my earlier naming suggestion: please ignore my previous `cov2_record_sidebar` comment.
I now think `custom_object_record_sidebar` would be more suitable for a public location key (clearer than internal cov2 naming).
@vijaytembugade  let me know if you see any concerns with this naming or have better suggestion before updating.

vijaytembugade: custom_object_record_sidebar, is good suggestion . But, developers should not be confused between cov1 and cov2, that is the only concern.

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -379,6 +393,47 @@ def invalid_location_flexible_error(location_options)
           validation_error
         end
 
+        def validate_object_types(manifest)
```
zendesk-punavwalke, line 398: Gate this validation behind the feature flag as well.
Right now, even when the flag is off, this still runs and can return object_types_required.
This required-check should only apply when the feature flag is enabled.

config/locales/en.yml
```diff
@@ -295,11 +298,22 @@ en:
...
             field_contains_invalid_keys: "%{field} contains invalid keys: %{keys}."
+            object_types_must_be_array: objectTypes must be an array of strings for
+              the %{location} location, but it was %{value}.
+            object_types_empty: objectTypes cannot be empty for the %{location} location.
+              Specify at least one object type.
+            object_types_invalid_entries: objectTypes must contain only non-empty
+              strings for the %{location} location.
+            object_types_not_supported: objectTypes is not supported for the %{location}
+              location.
+            object_types_required: objectTypes is required for the %{location} location.
```
zendesk-punavwalke, line 316: I don’t see the corresponding source translation entry in `config/locales/translations/zendesk_apps_support.yml`.
Please add it in the source translations file and regenerate en.yml

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -54,13 +54,14 @@ def collate_manifest_errors(package)
...
+        def validate_location(package, opts = {})
           manifest = package.manifest
           [
             missing_location_error(package),
-            invalid_location_error(package),
+            invalid_location_error(package, opts),
             invalid_v1_location(package),
-            location_framework_mismatch(manifest)
+            location_framework_mismatch(manifest),
+            validate_object_types(manifest, opts)
```
zendesk-punavwalke: Non-blocking: for readability, consider gating at call site `validate_object_types(manifest, opts) if flag instead` of entering the method and returning early.

zendesk-punavwalke (dismissed):

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -379,6 +393,49 @@ def invalid_location_flexible_error(location_options)
           validation_error
         end
 
+        def validate_object_types(manifest, opts = {})
+          return nil unless opts.fetch(:validate_custom_object_record_sidebar_location, false)
+
+          custom_object_record_sidebar = ZendeskAppsSupport::Location::CUSTOM_OBJECT_RECORD_SIDEBAR_LOCATION
+          manifest.location_options.each do |location_options|
+            object_types = location_options.object_types
+            next if location_options.location.nil?
```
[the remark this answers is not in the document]
vijaytembugade: Following the existing pattern.

config/locales/en.yml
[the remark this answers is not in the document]
vijaytembugade: This is autogenerated.

config/locales/en.yml
```diff
@@ -295,11 +298,22 @@ en:
             oauth_parameter_cannot_be_secure: oauth parameter cannot be set to be
               secure.
             invalid_url: '%{field} must be a valid URL, got "%{value}".'
+            password_parameter_type_deprecated: 'Password parameter type can no longer
+              be used. Use Secure settings instead. Learn more: %{link}.'
```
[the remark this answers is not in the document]
vijaytembugade: This is autogenerated.

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -329,6 +333,16 @@ def invalid_location_error(package)
...
             invalid_locations = package.manifest.unknown_locations(product.name)
+
+            unless validate_custom_object_record_sidebar
+              requested_locations = package.manifest.location_options
+                                           .select { |lo| lo.location&.product_code == product.code }
+                                           .map { |lo| lo.location&.name }
+              if requested_locations.include?(custom_object_record_sidebar)
+                invalid_locations << custom_object_record_sidebar
+              end
+            end
```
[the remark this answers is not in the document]
vijaytembugade: Duplicates can't occur here. unknown_locations returns names that don't exist in the Location registry (i.e., manifest_keys - registered_locations). Since custom_object_record_sidebar is a registered location (id: 27), it will never be in the unknown_locations set. The append only runs when the flag is off to treat an otherwise-valid location as invalid — the two sources are mutually exclusive by construction.

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -379,6 +393,49 @@ def invalid_location_flexible_error(location_options)
...
+            object_types = location_options.object_types
+            next if location_options.location.nil?
+
+            location_name = location_options.location.name
+
+            if object_types.nil?
+              if location_name == custom_object_record_sidebar
+                return ValidationError.new(:object_types_required,
+                                           location: location_name)
+              end
```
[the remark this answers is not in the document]
vijaytembugade: Updated the description

lib/zendesk_apps_support/validations/manifest.rb
```diff
@@ -379,6 +393,49 @@ def invalid_location_flexible_error(location_options)
           validation_error
         end
 
+        def validate_object_types(manifest, opts = {})
+          return nil unless opts.fetch(:validate_custom_object_record_sidebar_location, false)
```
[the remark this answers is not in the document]
vijaytembugade: Updated the description.

lib/zendesk_apps_support/package.rb
```diff
@@ -34,9 +34,11 @@ def initialize(dir, is_cached = true)
     def validate(options = {})
       marketplace = options.fetch(:marketplace, true)
       skip_marketplace_translations = options.fetch(:skip_marketplace_translations, false)
+      validate_custom_object_sidebar = options.fetch(:validate_custom_object_record_sidebar_location, false)
```
mmassaki: Consider naming the variable the same as the option for consistency.


```suggestion
      validate_custom_object_record_sidebar_location = options.fetch(:validate_custom_object_record_sidebar_location, false)
```

lib/zendesk_apps_support/package.rb
```diff
@@ -34,9 +34,11 @@ def initialize(dir, is_cached = true)
...
 
       errors = []
-      errors << Validations::Manifest.call(self)
+      manifest_opts = { validate_custom_object_record_sidebar_location: validate_custom_object_sidebar }
+      errors << Validations::Manifest.call(self, manifest_opts)
```
mmassaki: Consider using a keyword arg instead of a generic opts to simplify and make it easier to cleanup when removing the flag.


```suggestion
      errors << Validations::Manifest.call(self, validate_custom_object_record_sidebar_location: validate_custom_object_record_sidebar_location)
```

lib/zendesk_apps_support/validations/manifest.rb
mmassaki: +1 for this change.

By using a keyword arg in the methods in this class like https://github.com/zendesk/zendesk_apps_support/pull/428/changes#r3549503069, we can avoid duplicating the code to retrieve the option as well.

lib/zendesk_apps_support/location.rb
```diff
@@ -65,7 +67,9 @@ def self.all
       Location.new(id: 23, name: 'call_log_editor', product_code: Product::SELL.code, visible: true),
       Location.new(id: 24, name: 'email_editor', product_code: Product::SELL.code, visible: true),
       Location.new(id: 25, name: 'top_bar', product_code: Product::SELL.code, visible: true),
-      Location.new(id: 26, name: 'visit_editor', product_code: Product::SELL.code, visible: true)
+      Location.new(id: 26, name: 'visit_editor', product_code: Product::SELL.code, visible: true),
+      Location.new(id: 27, orderable: true, collapsible: true, name: 'custom_object_record_sidebar',
```
mmassaki: Consider using `CUSTOM_OBJECT_RECORD_SIDEBAR_LOCATION` for name if they are the same concept.

spec/validations/manifest_spec.rb
```diff
@@ -807,6 +807,210 @@ def create_package(parameter_hash)
...
+            location_name => location_opts
+          }
+        }
+      }
+    end
+
+    context 'when flag is enabled' do
+      def validate(package)
+        ZendeskAppsSupport::Validations::Manifest.call(package, enabled_opts)
+      end
```
mmassaki: Consider the following suggestions to improve readability:
1. clear description of the context to avoid ambiguity. e.g. "when validate_custom_object_record_sidebar_location is true".
2. replacing let blocks like `enabled_opts` and `disabled_opts` that are used in a single place by using literals or local variables.

```suggestion
    context 'when validate_custom_object_record_sidebar_location is true' do
      def validate(package)
        ZendeskAppsSupport::Validations::Manifest.call(package, validate_custom_object_record_sidebar_location: true)
      end
```

mmassaki (dismissed):

spec/validations/manifest_spec.rb
```diff
@@ -807,6 +807,207 @@ def create_package(parameter_hash)
...
+
+        it 'has an error when objectTypes is missing' do
+          @manifest_hash['location']['support'][location_name] = { 'url' => location_url }
+          create_package(@manifest_hash)
+          errors = validate(@package)
+          expect(errors.find { |e| e.key == :object_types_required }).not_to be_nil
+        end
+      end
+
+      context 'with supported location' do
```
zendesk-punavwalke: Please rename this context `from with supported location` to something explicit 
e.g `with non-custom_object_record_sidebar location`

spec/validations/manifest_spec.rb
```diff
@@ -807,6 +807,207 @@ def create_package(parameter_hash)
...
+
+        it 'does not validate objectTypes when missing' do
+          @manifest_hash['location']['support'][location_name] = { 'url' => location_url }
+          create_package(@manifest_hash)
+          errors = validate(@package)
+          expect(errors.find { |e| e.key == :object_types_required }).to be_nil
+        end
+      end
+
+      context 'with supported location' do
```
zendesk-punavwalke: same here as well and all the places where applicable

zendesk-punavwalke (approved):

mmassaki (approved):

